PR #134, opened 2015-01-08
Update date in index.html

alexspeller:
Now no one will ask if it's still active ;)


mixonic:
That's right! :-) :-)

JoshSmith:
:beers:

PR #138, opened 2015-01-13
Pass options from provider open() to popup open()

vincent99:
Adds passing the options from oauth provider's `.open()` method
to the popup window's `.open()` so that the caller can override
the width and height of the popup.  Currently there are options but
no way to actually set them from outside the library.

### DIFF
--- a/lib/torii/providers/facebook-oauth2.js
+++ b/lib/torii/providers/facebook-oauth2.js
@@ -19,8 +19,8 @@ export default Oauth2.extend({
     return this._super();
   }),
 
-  open: function() {
-    return this._super().then(function(authData){
+  open: function(options) {
+    return this._super(options).then(function(authData){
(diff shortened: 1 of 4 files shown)


vanstee:
Oh nice. I've needed this as well.

joostdevries:
:+1:

bagby:
I need this toooooo...

mixonic:
Ace, I am seeing the point now. We would need this to have a test in order to merge, and I would prefer that we not pass all options but have a nested object for popup-specific options.

wkoffel:
Is this still something that's in-flight?  The too-tiny oauth pop-up windows are definitely something we are looking to customize.

geoffroymontel:
Yes, I'm still waiting for being able to resize the OAuth popup window

PR #143, opened 2015-02-13
add azuread oauth2 provider

albertjan:


albertjan:
I haven't run the tests yet because spawn doesn't work on windows. I will try to look at that this weekend.

albertjan:
I will update this if and when #145 is pulled in. So I can change it so you can configure it to use openid connect.

mixonic:
can `azuread-oauth2` be rewritten as `azure-ad-oauth2`? Just to be consistent with the automatic camelizing rules?

Looks great! Thanks for the test!

mixonic:
@albertjan thanks for your patience :-p. This is good to go, except two small things:

* now that #145 is in, you wanted to make a change?
* `test/tests/unit/providers/google-oauth2-test.js` has an extra `w` at the bottom of the file.

Thanks for your patience, lets get it in.

albertjan:
Allright :smile: give me a minute or two. :trollface:

mixonic:
You da man. <3

albertjan:
Allright this is about it. I do think some documentation might be nice :smile:. And maybe sending a nonce with the request, but that opens up a whole new can of worms so I didn't.

mixonic:
@albertjan thanks sir!

albertjan:
Thank you! I will try and do some docs next week.

PR #145, opened 2015-02-19
use responsetype key for authdata

albertjan:
When using openid connect the responsecode is not in `code` but in `id_token`. Which is the same as the `responseType` property. This changes it so it uses the `responseType` value as key for getting a token or code from the `authData`.

Will update the azuread PR after this is pulled in.


mixonic:
@albertjan can you add a test for this behavior? It should be pretty straight ahead.

albertjan:
@mixonic I've added a test. Is there anything else you want me to change/add before you can pull this in?

mixonic:
@albertjan excellent!!! Thanks for the test and implementation :-)

PR #150, opened 2015-03-11
Added Stripe Connect Provider

OrKoN:
Please let me know if tests are need... The code is pretty straight-forward and works for in a test app. Also fixed a typo in the README


mixonic:
Very cool! A basic functionality test would indeed be appreciated. It is the only way we can be sure this doesn't break in the future.

OrKoN:
@mixonic I've added an integration test similar to the one for Github OAuth. Is it sufficient? 

P.S. found a bug in my initial pull request... :-( forgot to register the new provider

mixonic:
> P.S. found a bug in my initial pull request... :-( forgot to register the new provider

@OrKoN you're talking about the stripe registration? Looks like that is fixed here?

> I've added an integration test similar to the one for Github OAuth. Is it sufficient?

Well, this looks good to me. Are you willing trust that it is robust enough that a random committer doesn't break your app? This is the test, not some arbitrary bar by me :-)

looks good!

OrKoN:
@mixonic No, I meant the registration in the container in this file lib/torii/bootstrap/torii.js
This pull request contains the fixed version so it can be merged. All tests are passing... So I think it it should not break :-)

mixonic:
Thanks!

Issue #147, opened 2015-02-24
Ember-cli 0.2.x compatability


matthiasleitner:
Compatibility with the new ember-cli beta seems to be broken. Getting this error:

TypeError: Cannot read property 'env' of undefined
torii/ember-addon/index.js:15:48
at
  var config = this.project.config(this.app.env);

Not sure how this is related to changes like the addition of https://github.com/ember-cli/ember-cli/blob/master/ADDON_HOOKS.md#setuppreprocessorregistry

mixonic:
@rwjblue had some thoughts for us.

Frozenfire92:
This may be related to https://github.com/simplabs/ember-cli-simple-auth-torii/issues/7 which doesn't seem to be fixed

OrKoN:
@mixonic This happens to me still (version 0.3.2):

```
$ ember -v
version: 0.2.0
node: 0.12.0
npm: 2.7.0
```

and it throws this: 

```
Cannot read property 'env' of undefined
TypeError: Cannot read property 'env' of undefined
    at Class.treeForAddon (/tmp/stripe-test/node_modules/torii/ember-addon/index.js:26:46)
    at Class._treeFor (/tmp/stripe-test/node_modules/ember-cli/lib/models/addon.js:288:31)
    at Class.treeFor (/tmp/stripe-test/node_modules/ember-cli/lib/models/addon.js:258:19)
    at /tmp/stripe-test/node_modules/ember-cli/lib/models/addon.js:211:32
    at Array.map (native)
    at Class.eachAddonInvoke (/tmp/stripe-test/node_modules/ember-cli/lib/models/addon.js:209:22)
    at Class.treeFor (/tmp/stripe-test/node_modules/ember-cli/lib/models/addon.js:256:20)
    at EmberApp.<anonymous> (/tmp/stripe-test/node_modules/ember-cli/lib/broccoli/ember-app.js:307:20)
    at Array.map (native)
    at EmberApp.addonTreesFor (/tmp/stripe-test/node_modules/ember-cli/lib/broccoli/ember-app.js:305:30)
```

Updated ember-cli according to https://github.com/ember-cli/ember-cli/releases/tag/0.2.0 and created a new project:

```
ember new stripe-test 
cd stripe-test
ember install:addon ember-cli-simple-auth
npm install torii --save-dev
ember install:addon ember-cli-simple-auth-torii
ember s
```

PR #152, opened 2015-03-15
Fix incorrect `main` entry point script.

rwjblue:
The `ember-addon.main` path is still needed until https://github.com/ember-cli/ember-cli/pull/3545 is landed in a released version.

This fixes https://github.com/Vestorly/torii/issues/147.


mixonic:
Ah, make sense after reading through the diffs. Thanks very, very much for the dive sir.

PR #153, opened 2015-03-15
Allow usage from Ember CLI without built assets.

rwjblue:
* Serves `lib/torii` as the `addon` tree.
* Serves `lib/torii/initializers` as `app` tre at `app/initializers`.
* Removes `copy` task and `grunt-contrib-copy` from installed packages.

/cc @mixonic


rwjblue:
I did test against https://github.com/aptible/diesel.aptible.com (linked against this configuration) and all tests passed.  I would suggest taking it for a test drive as a sanity check...

mixonic:
Lookin' good so far. Will make a release shortly.

PR #167, opened 2015-04-06
Add google-oauth2-bearer provider

linstula:
This PR adds the google-oauth2-bearer provider, which allows Torii users to utilize the client-side OAuth flow outlined here: https://developers.google.com/accounts/docs/OAuth2UserAgent

I originally worked with @rwjblue to get this working in a dummy application before extracting it into a PR. There is quite a bit of duplication between the `google-oauth2-bearer` and `google-oauth2` providers and we discussed creating a mixin to remove the duplication but weren't sure what the desired approach would be. 

The primary difference with the client-side flow above and the server-side flow (https://developers.google.com/accounts/docs/OAuth2WebServer) is the client-side flow immediately grants the consuming application an `access_token` rather than granting an `authorization_code` which is then exchanged for an `access_token`.

Please let me know if there are any thoughts / desired changes on this PR. Thanks for the great library :+1:

### DIFF
--- a/lib/torii/bootstrap/torii.js
+++ b/lib/torii/bootstrap/torii.js
@@ -1,6 +1,7 @@
 import Torii from 'torii/torii';
 import LinkedInOauth2Provider from 'torii/providers/linked-in-oauth2';
 import GoogleOauth2Provider from 'torii/providers/google-oauth2';
+import GoogleOauth2BearerProvider from 'torii/providers/google-oauth2-bearer';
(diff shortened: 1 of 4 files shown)


rwjblue:
I teeter-tottered a bit mentally on the duplication initially too.  In the end it seems like the duplication is mostly non-logic so it doesn't seem like another layer (via a Mixin) that could be shared adds much value (IMHO).

:+1: - Looks good

mixonic:
Thank you for the excellent code and tests! :cake: Eager to try this out myself :-)

PR #177, opened 2015-05-05
Make redirect_uri optional for oauth2 providers

whatthewhat:
According to the spec the redirect_uri param is optional: https://tools.ietf.org/html/rfc6749#section-4.1.1

Currently a lot of tests fail because the GET params in the generated urls have switched places (`redirect_uri` has moved to the end of the url with the other optional params).

Before fixing the tests I'd like to get some feedback on these questions:

* Is the change legit? :smiley: 
* I would also remove the default value for `redirectUri`, but that could be a breaking change.. WDYT?
* Do any of the included adapters that inherit from `oauth2-code` need the old behaviour?

Thanks!

### DIFF
--- a/lib/torii/providers/oauth2-code.js
+++ b/lib/torii/providers/oauth2-code.js
@@ -42,7 +42,7 @@ var Oauth2 = Provider.extend({
    *
    * @property {array} requiredUrlParams
    */
-  requiredUrlParams: ['response_type', 'client_id', 'redirect_uri'],
+  requiredUrlParams: ['response_type', 'client_id'],
 
   /**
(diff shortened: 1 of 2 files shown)


mixonic:
@whatthewhat This seems legit, but I would like a real-world use case to consider. Are y'all doing this somewhere?

whatthewhat:
@mixonic Stripe does not require one (https://stripe.com/docs/connect/reference).
But after some thought I'm not sure it's worth introducing a potentially breaking change to other adapters  that inherit from `oauth2-code`..

I don't think there is a way to make the change to the stripe provider only though (or maybe there is a way to remove a value from `concatenatedProperties`?).

PR #179, opened 2015-05-06
Don't close the window immediately

### LINKED ISSUE #111: ember-cli torii facebook oauth2 Popup was closed or authorization was denied
So I have an application created in ember-cli. Torii is being injected properly on my route, and the Facebook Oauth2 provider launches properly and allows me to log in and receive an authData response.

However it gets a bit odd at that point, if my application is listed under the apps on my Facebook account (not on the developers site but, my personal Facebook as a user would access), launching the authentication using "this.get('torii').open" still triggers the popup, however it closes immediately and reports "Popup was closed or authorization was denied" as opposed to any auth data / any indication that I am signed in. 

The even stranger behaviour is, if I log into my Facebook and delete my app from the app settings and try to log in again, I will occasionally receive an authData response, otherwise it'll give me the usual "Popup was closed or authorization was denied" error. I am a bit new to ember-cli and Torii but I can't seem to pinpoint why I only sometimes receive the authData from Facebook properly and other times I don't. I feel like I might be doing something silly or overlooking something.

marcoow:
fixes #111

### DIFF
--- a/lib/torii/redirect-handler.js
+++ b/lib/torii/redirect-handler.js
@@ -19,7 +19,7 @@ var RedirectHandler = Ember.Object.extend({
     return new Ember.RSVP.Promise(function(resolve, reject){
       if (window.opener && window.opener.name === 'torii-opener') {
         postMessageFixed(window.opener, url);
-        window.close();
+        Ember.run.next(window, window.close);
       } else {
         reject('No window.opener');
       }


mixonic:
@marcoow This seems like it might mitigate but not solve the issue. Still happy to merge it though, pending a green build.

PR #182, opened 2015-05-10
wait a bit to receive postMessage in parent window. 

### LINKED ISSUE #111: ember-cli torii facebook oauth2 Popup was closed or authorization was denied

kbullaughey:
Instead of immediately failing when window is closed; wait for a few seconds to receive the message from postMessage; fixes #111.

I found this necessary at least on Chrome Canary and using the twitter provider.

### DIFF
--- a/lib/torii/services/popup.js
+++ b/lib/torii/services/popup.js
@@ -110,8 +110,8 @@ var Popup = Ember.Object.extend(Ember.Evented, {
         // If we don't receive a message before the timeout, we fail. Normally,
         // the message will be received and the window will close immediately.
         service.timeout = Ember.run.later(service, function() {
-          reject(new Error("Authorization never received authentication message."));
-        }, 5*1000);
+          reject(new Error("Popup was closed, authorization was denied, or a authentication message otherwise not received before the window closed."));
+        }, 100);
       });
 
       Ember.$(window).on('message.torii', function(event){
@@ -119,7 +119,9 @@ var Popup = Ember.Object.extend(Ember.Evented, {
         var toriiMessage = readToriiMessage(message);
         if (toriiMessage) {
           var data = parseMessage(toriiMessage, keys);
-          resolve(data);
+          Ember.run(function() {
+            resolve(data);
+          });
         }
       });
 
--- a/lib/torii/services/popup.js
+++ b/lib/torii/services/popup.js
@@ -107,8 +107,11 @@ var Popup = Ember.Object.extend(Ember.Evented, {
       }
 
       service.one('didClose', function(){
-        reject(new Error(
-          'Popup was closed or authorization was denied'));
+        // If we don't receive a message before the timeout, we fail. Normally,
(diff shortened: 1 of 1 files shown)


mixonic:
@kbullaughey Doesn't this only give the user five seconds to enter their credentials?

kbullaughey:
No, it gives them infinite time. The timeout only starts ticking when the window is closed. It give the browser time to communicate the `postMessage` call back to the parent. It seems some browsers kill off that message when the window is closed.

kbullaughey:
It also means that the redirect-handler is responsible for closing the window. Are there situations in which the Ember app would not be redirected to? For example, if authorization is rejected, does the window just close without the Ember app having a chance to do anything? In that case the timeout would start and the person would be left to wait it out, in which case the choice of 5 seconds might feel long.

mixonic:
ah, of course. So I actually see a different bug here that may be impacting things.

Line 122 in lib/torii/services/popup.js should be wrapped in an Ember runloop. As written, it may be causing an autorun which means the close event gets handled first.

Can you make a few quick changes:

* Change the error to: "Popup was closed, authorization was denied, or a authentication message otherwise not received before the window closed". It is verbose, but it encompassed the old message and the change you want to make here.
* Add a runloop around the `resolve` call.
* Make the timeout much shorter. I'm thinking 100ms. Even that is pretty huge- we're talking about a frame of JS here not a user interaction.

kbullaughey:
Okay, I've made those changes and checked that the new version works with my app and that tests pass, although I haven't looked into what the tests are or written any that specifically test this new functionality. I'm not very familiar with the torii code base yet.

mixonic:
Great! Thanks @kbullaughey :-)

fsmanuel:
@kbullaughey thanks

paulgbyrne:
Might be silly to mention this, but I had this problem for a half an hour only on Chrome, before realising that I had uBlocker installed, which was closing the popup.

davidbilling:
I still got this. Has this bug been reintroduced, can't see the fix in version 0.8.0, or is some new fix implemented?

PR #187, opened 2015-06-01
Use '_blank' and NOT name the window to work with chrome ios on iphone

anilmaurya:
For further detail, see this stackoverflow answer
http://stackoverflow.com/a/30226002/1970061


mixonic:
@anilmaurya can you also drop setting the name here:

* https://github.com/Vestorly/torii/blob/e4f369c6856c1d69eb24e0f9f9852ed72bbb5c3c/test/tests/unit/redirect-handler-test.js#L8

Looks great. I might even add a link to the stack overflow answer in a comment above the `_blank` name.

anilmaurya:
@mixonic updated PR, please check.

PR #191, opened 2015-06-17
Walk providers in a instance-initializer (instead of a regular initializer)

### LINKED ISSUE #190: Breaks with Ember Data 1.0.0-beta.19.2
**ember-data#1.0.0-beta.19.1**: `DEPRECATION: You tried to look up 'store:main', but this has been deprecated in favor of 'store:application'.`

**ember-data#1.0.0-beta.19.2**: `Uncaught Error: Attempting to inject an unknown injection: 'store:main'` raised from https://github.com/Vestorly/torii/blob/master/lib/torii/initializers/initialize-torii.js#L14 

---
torii: 0.3.5
ember: 2.0.0-beta.1

cibernox:
This fixes #190 

The main reason is that in modern versions of ember-data the store is registered using instance initializers, and since some providers are trying to inject the store, which is not defined at this point, then boom :boom:  with a quite cryptic error message.

This code will only work in versions of ember with instance initializers (1.12+). I'm not sure if bumping the version number to 0.4 and having support only 1.12, 1.13 and 2.0 is enough or we need to do something cleverer for backwards compatibility.

### DIFF
--- a/lib/torii/load-initializers.js
+++ b/lib/torii/load-initializers.js
@@ -1,10 +1,13 @@
 import loadInitializer from 'torii/lib/load-initializer';
+import loadInstanceInitializer from 'torii/lib/load-initializer';
 import initializeTorii from 'torii/initializers/initialize-torii';
 import initializeToriiCallback from 'torii/initializers/initialize-torii-callback';
 import initializeToriiSession from 'torii/initializers/initialize-torii-session';
+import walkProviders from 'torii/instance-initializers/walk-providers';
 
 export default function(){
   loadInitializer(initializeToriiCallback);
   loadInitializer(initializeTorii);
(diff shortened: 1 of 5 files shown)


lib/torii/load-initializers.js
```diff
@@ -1,10 +1,13 @@
 import loadInitializer from 'torii/lib/load-initializer';
+import loadInstanceInitializer from 'torii/lib/load-initializer';
```
bmac:
```js
import loadInstanceInitializer from 'torii/lib/load-instance-initializer';
```

cibernox:
Added compatibility matrix to the Readme

mixonic:
@cibernox Mucho gracias.

PR #195, opened 2015-06-26
Changed inter-window communication to use localStorage events

### LINKED ISSUE #193: Issue where callbacks aren't handled in Chrome and Firefox
I'm running into an issue where the Oauth callbacks aren't being handled by my application. After I sign in over stripe connect, instead of closing the popup window, my application just opens in the popup window.

screen shot 2015-06-24 at 2 29 45 pm

I am seeing this behavior in Firefox and Chrome on OSX but Safari works correctly. I have tried using torii 0.3.2 and 0.4.1, with ember 1.12.0 and ember 1.13.1, respectively. I have poked around in the Chrome debugger, and it doesn't seem like the window.postMessage code is running. Please let me know what information I can give you to help me figure out what's going on. Thanks in advance for your help.

aaronrenner:
This should fix #193. Any feedback and suggestions are welcome.

window.opener was being cleared after visiting a different host in Chrome 43 and Firefox 38. This change uses local storage events + unique popup ids to communicate between windows. Here’s an overview of how it works.

1. When popup.open is called, it generates a new torii popup id (uuid) and sets it as the popup window's name.
2. When RedirectHandler#handle is called in a window that has a valid torii popup id as its name, it stores the popup id and the current url of the popup window in local storage
3. When the new key/value is set in local storage, the popup service is notified via the browser's storage event. If event.key contains the popup id we’re looking for, it extracts the auth code from the url, resolves the promise, and removes the key from local storage.

Other notes

* We don't need to worry about opening popup with a name of `_blank` anymore. According to the commit log, the window was being named "_blank" because of an issue where window.opener wasn't being set in Chrome for iOS. Since we aren't using window.opener anymore, we should be able to set the window name to the popup id.

I have tested these changes in my application in the following browsers

* Chrome 43 on OSX 10.10
* Firefox 48 on OSX 10.10
* Safari 8.0.6 on OSX 10.10
* IE 11.0.9600 on Windows 8.1

Please let me know what you think, and if there are changes that need to be made.

### DIFF
--- a/lib/torii/services/popup.js
+++ b/lib/torii/services/popup.js
@@ -47,6 +47,7 @@ function parseMessage(url, keys){
 var Popup = Ember.Object.extend(Ember.Evented, {
 
   init: function(options){
+    this._super.apply(this, arguments);
     options = options || {};
     this.popupIdGenerator = options.popupIdGenerator || UUIDGenerator;
   },
@@ -92,7 +93,7 @@ var Popup = Ember.Object.extend(Ember.Evented, {
         }, 100);
       });
 
-      Ember.$(window).on('storage', function(event){
+      Ember.$(window).on('storage.torii', function(event){
         var storageEvent = event.originalEvent;
 
         var popupIdFromEvent = PopupIdSerializer.deserialize(storageEvent.key);
@@ -111,7 +112,7 @@ var Popup = Ember.Object.extend(Ember.Evented, {
       // didClose will reject this same promise, but it has already resolved.
       service.close();
       service.clearTimeout();
-      Ember.$(window).off('storage');
+      Ember.$(window).off('storage.torii');
     });
   },
 
--- a/lib/torii/services/popup.js
+++ b/lib/torii/services/popup.js
@@ -1,27 +1,8 @@
 import ParseQueryString from 'torii/lib/parse-query-string';
+import UUIDGenerator from 'torii/lib/uuid-generator';
+import PopupIdSerializer from 'torii/lib/popup-id-serializer';
 
 var on = Ember.on;
-
-var postMessageFixed;
-var postMessageDomain = window.location.protocol+'//'+window.location.host;
-var postMessagePrefix = "__torii_message:";
-// in IE11 window.attachEvent was removed.
-if (window.attachEvent) {
-  postMessageFixed = function postMessageFixed(win, data) {
-    win.postMessageWithFix(postMessagePrefix+data, postMessageDomain);
-  };
-  window.postMessageWithFix = function postMessageWithFix(data, domain) {
-    setTimeout(function(){
-      window.postMessage(data, domain);
-    }, 0);
-  };
-} else {
-  postMessageFixed = function postMessageFixed(win, data) {
-    win.postMessage(postMessagePrefix+data, postMessageDomain);
-  };
-}
-
-export {postMessageFixed};
 
 function stringifyOptions(options){
   var optionsStrings = [];
@@ -57,14 +38,6 @@ function prepareOptions(options){
   }, options);
 }
 
-function readToriiMessage(message){
-  if (message && typeof message === 'string' && message.indexOf(postMessagePrefix) === 0) {
-    return message.slice(postMessagePrefix.length);
-  }
-}
-
-export {readToriiMessage};
-
 function parseMessage(url, keys){
   var parser = new ParseQueryString(url, keys),
       data = parser.parse();
@@ -72,6 +45,11 @@ function parseMessage(url, keys){
 }
 
 var Popup = Ember.Object.extend(Ember.Evented, {
+
+  init: function(options){
+    options = options || {};
+    this.popupIdGenerator = options.popupIdGenerator || UUIDGenerator;
+  },
 
   // Open a popup window. Returns a promise that resolves or rejects
   // accoring to if the popup is redirected with arguments in the URL.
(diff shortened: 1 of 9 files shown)


mixonic:
Very interesting @aaronrenner!

> window.opener was being cleared after visiting a different host in Chrome 43 and Firefox 38

Do you suggest that this the buggy `opener` behavior was introduced in those browsers? Or is the reason for not having `window.opener` still unknown?

The caveats of localStorage would be:
* Private mode in Safari may not persist data across multiple tabs/windows.
* I think your code may handle this, but multiple windows opening providers may stomp on each other's localStorage keys

These do not seem like blockers. I would ideally like to know why opener is not a valid solution though.

aaronrenner:
@mixonic I think something changed 2 days ago with the window.opener behavior. It used to work just fine for me as written, but suddenly stopped working, even without code changes. I tried going back in my git history of my app and still couldn't get it to work. @bagby mentioned the same behavior in issue #193.

Multiple popup windows shouldn't stomp on each other because they are named `torii-popup:<uuid>`.

I just tested in safari private mode, and you're right, it doesn't work. Maybe we have to do a combination of both strategies, `window.opener` and local storage.

mixonic:
Great. At a glance this is looking excellent. I'll give it a deep look later today or tomorrow morning and lets get it merged and into a release. Thanks for catching this issue and putting together such a thorough PR.

bagby:
Any movement on this? It looks good to me, dunno if you've had a chance to look at it @mixonic.

whatthewhat:
Confirming that this branch fixes the issue (still not really sure why only `stripe-connect` is broken..)

lib/torii/services/popup.js
```diff
@@ -73,6 +46,11 @@ function parseMessage(url, keys){
 
 var Popup = Ember.Object.extend(Ember.Evented, {
 
+  init: function(options){
```
mixonic: An `this._super.apply(this, arguments);` at the start of `init` would be ideal

lib/torii/services/popup.js
```diff
@@ -114,11 +92,13 @@ var Popup = Ember.Object.extend(Ember.Evented, {
         }, 100);
       });
 
-      Ember.$(window).on('message.torii', function(event){
-        var message = event.originalEvent.data;
-        var toriiMessage = readToriiMessage(message);
-        if (toriiMessage) {
-          var data = parseMessage(toriiMessage, keys);
+      Ember.$(window).on('storage', function(event){
```
mixonic: Using `storage.torii` and `off('storage.torii')` would ensure we don't tear down anyone else's listener

mixonic:
@aaronrenner this looks great. If you can make the two small changes I've called out I will happily merge and make a release later today. Thanks :-)

bagby:
*poke* @aaronrenner 

sorry, we need this :smile:

aaronrenner:
Sorry guys. I was traveling yesterday. I'll make those changes this morning.

aaronrenner:
@mixonic I just pushed the changes you requested.

mixonic:
@aaronrenner thanks for this excellent work sir. Very appreciated!

OrKoN:
@aaronrenner @mixonic Were there any interface changes with this PR? I have updated my app to v0.5.0 and now the popup remains opened after authentication on Github and the promise is never resolved:

```
  this.get('torii').open('github-oauth2').then((authorization) => {
        Ember.$.post(window.ENV.API_HOST + '/auth/github', authorization)
        .then(this.onLogin.bind(this));
      }, () => {
        this.send('hideLoading');
      });
```

It does not work for me neither in FF nor in Chrome.

mixonic:
@OrKoN seems like a regression.

OrKoN:
@mixonic I see that the popup service wait for `Ember.$(window).on('storage.torii')` but nothing is triggering it. So I traced it down to the RedirectHandler where windowObject.name is read. But the window name is always empty and I guess it should contain the popup id?! Found this too: https://code.google.com/p/chromium/issues/detail?id=504164 seems to be causing what I see.

john-griffin:
I'm also getting the popup remaining open and the promise never resolving with `facebook-oauth2` on Chrome. This only happens on fresh logins where there is no existing Facebook session.

mixonic:
@OrKoN @john-griffin I've tried the demos on Chrome 44 and I'm not seeing any unexpected behavior. You can give me further steps to reproduce?

mixonic:
Ah, in FireFox 38 I have the window staying open. Would love a repro on chrome.

OrKoN:
@mixonic I have not figured out why it happens yet. I tested in an ember-cli app and it worked fine. But my app does not use ember-cli and although it seems that I initialize torii correctly, I have this problem in both FF and Chrome. So far I am staying on v0.4.1 (works fine). 

The code of my app is like this:

```JavaScript
import Ember from 'ember';
import toriiInitializers from 'torii/load-initializers';
toriiInitializers();

import ApplicationAdapter from 'torii-adapters/application';
import StackoverflowProvider from './torii-providers/stackoverflow';

Ember.Application.initializer({
  name: 'torii-adapters',
  initialize: function(container, application) {
    application.register('torii-adapter:application', ApplicationAdapter);
    application.register('torii-provider:stackoverflow', StackoverflowProvider);
    application.inject('controller', 'torii', 'torii:main');
  }
});

```

and the config is like:

```JavaScript
window.ENV.torii = {
  providers: {
    'github-oauth2': {
      apiKey: 'XXXXX',
      redirectUri: window.ENV.APP_HOST,
      scope: 'user:email'
    },
    stackoverflow: {
      apiKey: 'XXXX',
      redirectUri: window.ENV.APP_HOST
    }
  }
};

```

mixonic:
@OrKoN "I have this problem in both FF and Chrome", but what version of Chrome?

OrKoN:
@mixonic It's 43.0.2357.132

mixonic:
@aaronrenner it appears `window.name` is not preserved across domains :-(

mixonic:
ah right this is old knowledge and I should have caught it earlier. Facebook sets the window.name itself, which smashes the one we have set.

working on a patch to just use a global key in localStorage.

PR #196, opened 2015-06-26
Added always_prompt option for stripe connect

aaronrenner:
This pull request allows the developer to configure the always_prompt option for stripe connect.


mixonic:
Thank you!

PR #197, opened 2015-07-10
Remove use of window.name as a channel for localStorage messages

mixonic:
window.name was used to decide what item key would be used for a popup's message to the parent. This is nor reliable, as some browsers leak the fact that Facebook (and other popups) may set a window name that destroys the one set by Torii.

Instead, this patch uses a global key set by torii, identifying a "current channel" that expects to receive the message. This is not perfect, as a window opening a torii request after an different window
may receive the completed auth from the window of the first request. However in a practical sense it means the last window starting auth will receive the messages from auth popups, and old windows that may have started auth will receive none.

See https://github.com/Vestorly/torii/pull/195#issuecomment-119669300

### DIFF
--- a/lib/torii/redirect-handler.js
+++ b/lib/torii/redirect-handler.js
@@ -7,7 +7,7 @@
  */
 
 import PopupIdSerializer from "./lib/popup-id-serializer";
-
+import { CURRENT_REQUEST_KEY } from "./services/popup";
 
 var RedirectHandler = Ember.Object.extend({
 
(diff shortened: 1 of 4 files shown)


mixonic:
@aaronrenner could use a sanity check.

@OrKoN if you can give this a real-world smoke test that would be awesome. I believe it addresses your issues in Chrome and FF.

OrKoN:
@mixonic Thanks a lot. Worked in my case.

lib/torii/redirect-handler.js
```diff
@@ -19,13 +19,11 @@ var RedirectHandler = Ember.Object.extend({
     var windowObject = this.windowObject;
 
     return new Ember.RSVP.Promise(function(resolve, reject){
-      var popupId = PopupIdSerializer.deserialize(windowObject.name);
-
-      if (popupId){
-        var key = PopupIdSerializer.serialize(popupId);
+      var pendingRequestKey = windowObject.localStorage.getItem(CURRENT_REQUEST_KEY);
```
aaronrenner: @mixonic It looks like this approach might only allow the last opened popup to be resolved. I was trying to address the use case where a user has 2 torii popups open (like they clicked the login button twice and it created 2 popups).

mixonic: With this code:

* If a parent page opened two popups, either popup could login or cancel.
* If two parent pages opened a popup each, only the second parent page would handle the login or cancel.

IMO this edge-case seems to be ok?

aaronrenner: I agree, that seems ok for now.

PR #198, opened 2015-07-10
facebook-connect - Add support for return scopes

john-griffin:
Adds a configuration option to facebook-connect provider to return the granted scopes back with the response data. This is useful when trying to determine if a user omitted any required scopes during the sign in process.

Defaults to off and is only added to the normalized response when configured.


mixonic:
@john-griffin this looks great- Can you add a test for it?

john-griffin:
@mixonic good point, updated with a test.

mixonic:
:+1: Thanks!

PR #199, opened 2015-07-13
facebook-connect - Add support for auth_type

john-griffin:
Adds the ability to set the `auth_type` when connecting to Facebook. Currently
Facebook only supports `rerequest` which allows for requesting declined 
permissions.

This is not suitable for global configuration because you may not always
want to rerequest permissions but do it based on other conditions. As
such its passed through to the open function in options. It defaults
to not being set (aka the standard facebook-connect auth_type).

### DIFF
--- a/lib/torii/providers/facebook-connect.js
+++ b/lib/torii/providers/facebook-connect.js
@@ -34,7 +34,7 @@ function fbLoad(settings){
   return fbPromise;
 }
 
-function fbLogin(scope, returnScopes){
+function fbLogin(scope, returnScopes, authType){
   return new Ember.RSVP.Promise(function(resolve, reject){
     FB.login(function(response){
       if (response.authResponse) {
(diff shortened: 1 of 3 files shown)


lib/torii/providers/facebook-connect.js
```diff
@@ -73,12 +73,14 @@ var Facebook = Provider.extend({
   // API:
   //
   open: function(){
+    var options = arguments[0] === undefined ? {} : arguments[0];
```
mixonic: Why use `arguments[0]` instead of just accepting `options` as a param?

john-griffin: Converted this feature from my app's custom provider which uses ES6 to default the `options` to `{}`. Have updated and rebased. Thanks!

john-griffin:
@mixonic I updated that options param.

mixonic:
Thank you @john-griffin, apologies for the delay.

PR #200, opened 2015-07-14
authenticatedRoute Router DSL

taras:
This PR is a follow up to a conversation that @mixonic and I had via Slack about eliminating the need for adding a route mixin to make routes ensure they have authentication. Please, consider this a rough draft. 

This PR provides function called `authenticatedRoute` that can be used in Router map to mark specific routes as requiring authentication. Here is an example:

```js
import Ember from 'ember';
import config from './config/environment';
var Router = Ember.Router.extend({
  location: config.locationType
});

export default Router.map(function() {
  this.authenticatedRoute('add');
});
```

When at least one `authenticatedRoute` route is present, the app will automatically attempt to `checkLogin` which will call `session.fetch`. To reduce boilerplate, I added a `sessionProvider` property to Torii config which allows to set which session provider should be fetched: ie. `this.session.fetch(config.sessionProvider)`

What do you think about this?

Questions:
~~1. Currently, the user has to overwrite `accessDenied` on each route. Should we take that from ApplicationRoute? or maybe send a `accessDenied` action?~~
~~2. What do you think about `sessionProvider`?~~

## TODO

- [x] `authenticatedRoute` function 
- [x] Test mixin is not added if sessionName is not specified
- [x] Test mixin is added when sessionName is specified
- [x] Test `checkLogin` is not called when there are no `authentictedRoutes`
- [x] Test `checkLogin` is called with `authenticatedRoutes` present
- [x] Test `accessDenied` hook is called when access is denied
- [x] Test session is retrieved
- [x] Test session.fetchDefaultProvider
- [x] Add authenticatedRoute Router DSL Extension
- [x] Add test for Router DSL Extension
- [x] Ensure accessDenied action is sent to transition
- [x] Add documentation to README
- [x] Remove code targeting < Ember 1.12
- [x] Use container lookup to get the application adapter instead of `sessionProvider` key

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,10 +58,12 @@ export default Ember.Route.extend({
 torii.open('facebook') -> #open hook on the facebook provider -> returned authorization
 ```
 
-This is authentication only against a provider. If your application provides
-an **adapter**, then Torii can also perform **session management** via the
-`session` property, injected onto routes and controllers. This example uses
-Facebook's OAuth 2.0 API directly to fetch an authorization code.
+## Session Management
+
+Torii can perform **session management** via the `session` service, injected onto
+routes and controllers. You can activate *session management* by specifying `sessionServiceName` in your `config/environment.js` and providing an **adapter** which Torii will use to extract session information from a new opened authorization.
+
+This example uses Facebook's OAuth 2.0 API directly to fetch an authorization code.
 
 ```JavaScript
 /* jshint node: true */
@@ -144,6 +146,89 @@ session.open('facebook') -> #open hook on the facebook provider -> #open hook on
 ```
 
 Note that the adapter section is left entirely to your application.
+
+## Router DSL
+
+Torii includes a mechanism for specifying authenticated routes via the `Router.map` and a common implementation of authentication flow for your application.
+
+The authentication flow is as follows:
+
+1. In Application Route, check if the user is logged in by calling ApplicationRoute#checkLogin which calls `this.session.fetch()`.
+2. When entering an authenticated route,
+    Attempt to authenticate by calling `this.session.fetch()`
+      If successful,
+        allow the transition to finish
+      otherwise,
+        interrupt the transition and send "accessDenied" action
+
+This example uses Facebook's OAuth 2.0 API and the authenticatedRoute DSL.
+
+```JavaScript
+/* jshint node: true */
+// config/environment.js
+module.exports = function(environment) {
+  var ENV = {
+    /* ... */
+    torii: {
+      // a 'session' property will be injected on routes and controllers
+      sessionServiceName: 'session',
+      providers: {
+        'facebook-oauth2': {
+          apiKey:      'facebook-app-id',
+          redirectUri: '/my-custom-landing-uri' // default is the current URL
+        }
+      }
+    }
+  };
+  return ENV;
+};
+```
+
+```JavaScript
+// app/router.js
+Router.map(function(){
+    this.authenticatedRoute('my-account');
+    this.route('login');
+});
+```
+
+```JavaScript
+// app/routes/application.js
+export default Ember.Route.extend({
+  actions: {
+    accessDenied: function() {
+      this.transitionTo('login');
+    }
+  }
+});
+```
+
+```JavaScript
+// app/torii-adapters/application.js
+export default Ember.Object.extend({
+  open: function(authentication){
+    var authorizationCode = authentication.authorizationCode;
+    return new Ember.RSVP.Promise(function(resolve, reject){
+      Ember.$.ajax({
+        url: 'api/session',
+        data: { 'facebook-auth-code': authorizationCode },
+        dataType: 'json',
+        success: Ember.run.bind(null, resolve),
+        error: Ember.run.bind(null, reject)
+      });
+    }).then(function(user){
+      // The returned object is merged onto the session (basically). Here
+      // you may also want to persist the new session with cookies or via
+      // localStorage.
+      return {
+        currentUser: user
+      };
+    });
+  }
+});
+```
+
+The session will automatically be populated if the user is logged in, otherwise the user will be redirected to the login page.
 
 ## Using Torii
 
--- a/bower.json
+++ b/bower.json
@@ -3,9 +3,18 @@
   "version": "0.5.1",
   "description": "A set of clean abstractions for authentication in Ember.js",
   "authors": [
-    {"name": "Matt Beale", "email": "matt.beale@madhatted.com"},
-    {"name": "Cory Forsyth", "email": "cory.forsyth@gmail.com"},
-    {"name": "Vestorly", "email": "hello@vestorly.com"}
+    {
+      "name": "Matt Beale",
+      "email": "matt.beale@madhatted.com"
+    },
+    {
+      "name": "Cory Forsyth",
+      "email": "cory.forsyth@gmail.com"
+    },
+    {
+      "name": "Vestorly",
+      "email": "hello@vestorly.com"
+    }
   ],
   "homepage": "http://vestorly.github.io/torii/",
   "keywords": [
@@ -40,14 +49,13 @@
     "url": "git://github.com/Vestorly/torii.git"
   },
   "dependencies": {
-    "ember": "^1.3.0"
+    "ember": ">= 1.8.1"
   },
   "devDependencies": {
-    "handlebars": "^1.3.0",
     "jquery": ">= 1.9.1",
     "qunit": "^1.12.0",
     "ember-qunit": "~0.1.5",
-    "ember-resolver": "git://github.com/stefanpenner/ember-jj-abrams-resolver.git#master",
+    "ember-resolver": "ember-cli/ember-resolver#v0.1.21",
     "loader": "git://github.com/stefanpenner/loader.js#1.0.0"
   }
 }
--- a/ember-addon/index.js
+++ b/ember-addon/index.js
@@ -10,11 +10,19 @@ module.exports = {
 
   treeForApp: function treeForApp(tree) {
     var Funnel = require('broccoli-funnel');
+    var mergeTrees = require('broccoli-merge-trees');
 
-    return new Funnel(tree, {
+    var initializers = new Funnel(tree, {
       srcDir: 'initializers',
       destDir: 'initializers'
     });
+
+    var instanceInitializers = new Funnel(tree, {
+      srcDir: 'instance-initializers',
+      destDir: 'instance-initializers'
+    })
+
+    return mergeTrees([initializers, instanceInitializers]);
   },
 
   treeForAddon: function treeForAddon() {
--- a/lib/torii/router-dsl-ext.js
+++ b/lib/torii/router-dsl-ext.js
@@ -1,0 +1,17 @@
+var Router = Ember.Router;
+var proto = Ember.RouterDSL.prototype;
+
+var currentMap = null;
+
+proto.authenticatedRoute = function() {
+  this.route.apply(this, arguments);
+  currentMap.push(arguments[0]);
+};
+
+Router.reopen({
+  _initRouterJs: function() {
+    currentMap = [];
+    this._super();
+    this.router.authenticatedRoutes = currentMap;
+  }
+});
--- a/lib/torii/routing/authenticated-route-mixin.js
+++ b/lib/torii/routing/authenticated-route-mixin.js
@@ -1,0 +1,32 @@
+export default Ember.Mixin.create({
+  beforeModel: function (transition) {
+    var route = this;
+    var superBefore = this._super.apply(this, arguments);
+    if (superBefore && superBefore.then) {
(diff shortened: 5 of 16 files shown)


mixonic:
> 1. Currently, the user has to overwrite accessDenied on each route. Should we take that from ApplicationRoute? or maybe send a accessDenied action?

Definitely should be an action so you can implement once on the application route, and so that an error is raise (action not handled) when no behavior is specified.

lib/torii/initializers/initialize-torii-routing.js
```diff
@@ -0,0 +1,9 @@
+import Ember from 'ember';
+import ToriiAuthenticatedRouteMixin from 'torii/routing/authenticated-route-mixin';
+
+export default {
+  name: 'torii-routing',
+  initialize: function(){
+    Ember.Route.reopen(ToriiAuthenticatedRouteMixin);
```
mixonic: Like the session service, this should be based on if `config.sessionServiceName` is actually present.

See https://github.com/Vestorly/torii/blob/master/lib/torii/initializers/initialize-torii-session.js#L9

lib/torii/routing.js
```diff
@@ -0,0 +1,12 @@
+import Ember from 'ember';
+
+export let authenticatedRoutes = [];
+
+export default function authenticatedRoute(dsl, name) {
+  var args = [].slice.call(arguments, 2);
```
mixonic: ES2015 baby:

```
function authenticatedRoute(dsl, ...dslArgs) {
  dsl.route.apply(dsl, ...dslArgs);
  authenticatedRoutes.push(dslArgs[0]);
}
```

lib/torii/routing/authenticated-route-mixin.js
```diff
@@ -0,0 +1,44 @@
+import Ember from 'ember';
+import {
+  authenticatedRoutes
+} from 'torii/routing';
+import configuration from 'torii/configuration';
+
+let isEmpty = Ember.isEmpty;
+
+export default Ember.Mixin.create({
+  beforeModel(transition) {
```
mixonic: Should call super:

```
let superBefore = this._super(...arguments);
if (superBefore && superBefore.then) {
  return superBefore.then(() => {
    return authBeforFilter();
  });
} else {
  return authBeforeFilter();
}
```

It is unlikely any other mixin was applied, but not impossible.

lib/torii/routing.js
taras: Nice!

taras:
@mixonic I applied your suggestions except for ES2015 in *lib/torii/routing.js*. I don't think the mechanism that you use to transpile the assets supports arguments destructuring.

taras:
What do you think about cleaning up the authenticatedRoutes global? How should it be done? When?

If you think that this could make it into Torii then I'll start working on tests. Please, let me know.

taras:
@mixonic I don't know how to make Application.Router trigger map when app is started. I started working on it but I'm stuck.

Here is what I have so far https://github.com/Vestorly/torii/pull/200/files#diff-82ed7f075f489a2eae962cb15853f3a6R23

Any idea?

taras:
I figured out the problem with `Router.map` not being called. Moving on to other authenticated route logic tests.

taras:
It's working :) and tests are done. Could use a few more tests - I think.

I think adding the DSL could be pretty easy. I can look into doing that over the weekend.

test/tests/acceptance/routing-test.js
```diff
@@ -0,0 +1,129 @@
...
+    originalSessionServiceName = configuration.sessionServiceName;
+    delete configuration.sessionServiceName;
+    originalSessionProvider = configuration.sessionProvider;
+    delete configuration.sessionProvider;
+  },
+
+  teardown: function(){
+    Ember.run(app, 'destroy');
+    configuration.sessionServiceName = originalSessionServiceName;
+    configuration.sessionProvider = originalSessionProvider;
```
bantic: should teardown call `resetAuthenticatedRoutes`?

taras: Hey Cori, I'm sorry, I forgot to push my last changes.

In the new code, teardown is not necessary because authenticatedRoutes are automatically turndown when the Router instance is torn down.

package.json
```diff
@@ -16,6 +16,7 @@
     "main": "ember-addon/index.js"
   },
   "devDependencies": {
+    "broccoli-merge-trees": "^0.2.2",
```
bantic: This is a dependency, right? Not dev.

taras: Yes, you're right. I'll change it.

test/tests/unit/routing/application-route-mixin-test.js
```diff
@@ -0,0 +1,71 @@
...
+        fetchDefaultProviderCalled = true;
+        return Ember.RSVP.reject('no session is available');
+      }
+    }
+  }).create();
+
+  route.checkLogin()
+    .then(function(){
+      assert.ok(fetchDefaultProviderCalled, 'fetch default provider was called');
+      assert.ok('successful callback inspite of rejection');
```
bantic: nitpick: "in spite" is 2 words

test/tests/unit/routing/authenticated-route-mixin-test.js
```diff
@@ -0,0 +1,133 @@
...
+    }
+  });
+
+  return route.authenticate()
+    .then(function(){
+      assert.ok(true);
+    })
+});
+
+test('unattempted authentication results in authentication attempt', function(assert){
```
bantic: test should say "attempted authentication" right?

bantic:
@taras This looks great. It would be great to have a meaningful error message if a consumer uses this without having set `config.sessionProvider` — if I'm reading it correctly they'll get an undefined function error instead.

What do you have in mind for the DSL? Something like a function-returning `map` function from torii?
```
import toriiMap from 'torii/routing/map';

Router.map(toriiMap(function() {
  this.authenticatedRoute('secret', ...);
  this.route('other', ...);
});
```
?

bantic:
Are you still planning to make `accessDenied` an action that can be bubbled up?

taras:
Yes, I'm not sure if you saw the last commit, but it has the Router DSL extension.

taras:
In the latest code, the DSL is this

```
import Ember from 'ember';
import config from './config/environment';
var Router = Ember.Router.extend({
  location: config.locationType
});

export default Router.map(function() {
  this.authenticatedRoute('add');
});
```

bantic:
@taras Thanks! Aside from getting tests for the new aspects, this looks great. I'll leave a final :+1: to @mixonic since he has more context on your earlier discussions. Could you also add a checkbox to this PR desc for updating the README with the new functionality?

test/tests/unit/routing/authenticated-route-mixin-test.js
taras: No, this is testing that if we haven't tried to authenticate then we should try. Basically, the logic is:

* Not tried authenticating -> authenticate
* Previously successful authentication -> successful promise resolution
* Previously failed authentication -> failed promise returned with accessDenied action sent

bantic: Ah, ok. What I am seeing is:
  * call `authenticated` on a route
  * resolve its promise and assert that `fetchDefaultProvider` was called

Maybe the appropriate test name is "attempting authentication calls `fetchDefaultProvider`"? It's just the test title that I find a little unclear, for this one: `'unattempted authentication results in authentication attempt'`

taras: That is clearer :) Thank you

lib/torii/router-dsl-ext.js
```diff
@@ -0,0 +1,18 @@
...
+var proto = Ember.RouterDSL.prototype;
+
+var currentMap = null;
+
+proto.authenticatedRoute = function() {
+  this.route.apply(this, arguments);
+  currentMap.push(arguments[0]);
+};
+
+// 1.10 and above
```
mixonic: 0.4.x is already >= 1.12 :+1:

mixonic: I'm not crazy about this using a private API tho. The benefit of the import style was it being very future-safe.

If both liquid-fire and this library extend the DSL, we should probably attempt a public API.

lib/torii/session.js
```diff
@@ -93,8 +94,14 @@ var Session = Ember.ObjectProxy.extend({
       sm.send('failClose', error);
       return Ember.RSVP.reject(error);
     });
-  }
+  },
 
+  fetchDefaultProvider: function() {
+    Ember.assert(
+      'You must specify a session provider in your torii configuration',
+      configuration.sessionProvider);
```
mixonic: Can we follow Ember conventions here, and just attempt to load the `application` provider? Then perhaps throw if it is not present?

It would remove the need for a `sessionProvider` config value. Instead, just like with your ember-data adapters, you would create an application adapter and export.

lib/torii/router-dsl-ext.js
taras: > If both liquid-fire and this library extend the DSL, we should probably attempt a public API in Ember soon.

Agree and API for this would be great.

lib/torii/session.js
taras: > Can we follow Ember conventions here, and just attempt to load the application provider? Then perhaps throw if it is not present?

Ok, will do.

taras:
@mixonic I squashed the original 22 commits into 1, rebased against master and pushed 3 new commits.

I removed `sessionProvider` and now just using `session.fetch` which should automatically use the application adapter.

taras:
@mixonic @bantic I changed the README to information about how to use the `authenticatedRoute` DSL

README.md
```diff
@@ -145,6 +148,89 @@ session.open('facebook') -> #open hook on the facebook provider -> #open hook on
 
 Note that the adapter section is left entirely to your application.
 
+## authenticatedRoute Router DSL
```
mixonic: Just `Router DSL` seems fine, or a title less tied to implementation and more to intent. Not letting it stop me from merging though.

bower.json
```diff
@@ -40,14 +49,13 @@
     "url": "git://github.com/Vestorly/torii.git"
   },
   "dependencies": {
-    "ember": "^1.3.0"
+    "ember": ">= 1.8.1 < 2.0.0"
```
mixonic: Should actually be `>= 1.12`. I do not suggest `< 2.0`. Is there a problem with 2.0?

ember-addon/index.js
```diff
@@ -10,11 +10,19 @@ module.exports = {
...
 
-    return new Funnel(tree, {
+    var initializers = new Funnel(tree, {
       srcDir: 'initializers',
       destDir: 'initializers'
     });
+
+    var instanceInitializers = new Funnel(tree, {
+      srcDir: 'instance-initializers',
+      destDir: 'instance-initializers'
```
mixonic: :+1:

lib/torii/services/torii-session.js
```diff
@@ -1,4 +1,5 @@
 import createStateMachine from 'torii/session/state-machine';
+import configuration from 'torii/configuration';
```
mixonic: no longer needed?

bower.json
taras: I haven't tried it in 2.0 - I'm not sure. (would love to move this addon to an EmberCLI addon, we'd get 2.0 testing for free)

lib/torii/services/torii-session.js
taras: yep, removed

test/helpers/start-app.js
```diff
@@ -14,10 +15,13 @@ function startApp(attrs) {
   }, attrs); // but you can override;
 
   var Application = Ember.Application.extend();
-  Application.Router = Ember.Router.extend({
+
+  (Router || Ember.Router.extend()).reopen({
```
mixonic: If `Router` is not present, this will not set `Application.Router`. Seems like you want:

```js
Router = Router || Ember.Router.extend();
Router.reopen({
```

README.md
```diff
@@ -58,10 +58,13 @@ export default Ember.Route.extend({
...
 ```
 
-This is authentication only against a provider. If your application provides
-an **adapter**, then Torii can also perform **session management** via the
-`session` property, injected onto routes and controllers. This example uses
-Facebook's OAuth 2.0 API directly to fetch an authorization code.
+## Session Management
+
+Torii can perform **session management** via the `session` property, injected onto
+routes and controllers. You can activate *session management* by providing an **adapter**
```
mixonic: This is not actually accurate. Session management is activated by setting `sessionServiceName`, not by creating a provider.

test/helpers/start-app.js
taras: done

mixonic:
Taras, thank you for your patience and effort. This is a great addition, and I'm excited to exercise it in the real world.

PR #201, opened 2015-07-15
Add a test helper that stubs a valid torii session.

jagthedrummer:
Based on the suggestion from @bantic in this comment https://github.com/Vestorly/torii/issues/103#issuecomment-97807965 I've added a `stubValidSession` helper method.

You can access it with this import:

```javascript
import {stubValidSession} from 'torii/helpers/testing';
```

And then call it passing in the test `application` and the object that should be assigned as the `currentUser` in the session.

```javascript
stubValidSession(application, {handle : 'testguy', uid : 'xyz'});
```

An entire acceptance test might look like this:

```javascript
import Ember from 'ember';
import {
  module,
  test
} from 'qunit';
import startApp from 'nerdbeers-ember/tests/helpers/start-app';
import {stubValidSession} from 'torii/helpers/testing';

var application;

module('Acceptance: Index', {
  beforeEach: function() {
    application = startApp();
  },

  afterEach: function() {
    Ember.run(application, 'destroy');
  }
});

test('visiting / when signed out should prompt you to sign in', function(assert) {
  visit('/');

  andThen(function() {
    assert.equal(find('button:contains(Sign in with github)').length, 1);
  });
});

test('visiting / when signed in should not prompt you to sign in', function(assert) {
  stubValidSession(application, {handle : 'testguy', uid : 'xyz'});
  visit('/');

  andThen(function() {
    assert.equal(find('button:contains(Sign in with github)').length, 0);
    assert.equal(find('p:contains(Signed in as : testguy)').length, 1);
  });
});

```

With `index.hbs`

```handlebars
{{#if session.isAuthenticated}}
  <p>Signed in as : {{session.currentUser.handle}}</p>
{{else}}
   <button {{action "signIn" "github"}}>Sign in with github</button>
{{/if}}
```


mixonic:
I love the API, but I am conflicted about the wisdom in using a private API to access `_container_`.

It should probably also be part of the `test-support` tree. As demonstrated [here in ember-cli-aptible-shared](https://github.com/aptible/ember-cli-aptible-shared/tree/master/test-support), the test-support tree does not ship with app code, only into the testing assets.

jagthedrummer:
Ah-ha! I had a feeling that it wasn't in the right place for being only a test helper. I'll move it into `test-support`.

Regarding the `_container_` issue, do you have a suggestion for how to approach it a different way? If so I'm happy to change it up.

jagthedrummer:
Is there some trick to using stuff that lives in `test-support`? I've moved the helper file to live in `torii/test-support/helpers/testing.js`, but I can't seem to figure out a valid way to `import` it in the test.

I've tried all of these:

```javascript
import {stubValidSession} from 'torii/test-support/helpers/testing.js';
import {stubValidSession} from 'test-support/helpers/testing.js';
import {stubValidSession} from '../../../test-support/helpers/testing.js';
```

With any of them I get an error when running the tests that looks like:

```
Uncaught Error: Could not find module torii/test-support/helpers/testing.js
```

bantic:
@jagthedrummer You can see what AMD modules are defined by looking at the global `requirejs.entries`. You're talking about importing it from a consuming app's tests, right?

It may be that the test-support tree is mixed in with the `"<app-name>/tests/"` tree, so you would import it like: `import {stubValidSession} from "../<torii-test-helper-module-name>"` (in an acceptance test at, say, `"tests/acceptance/my-test.js"`).

jagthedrummer:
Well, for now I'm just trying to import it into the acceptance test in the torii test suite and not having any luck. When I inspect `requirejs.entries` in the test runner browser window I don't find my helper class in there at all if it lives in `torii/test-support`. If it's either in `torii/test` or `torii/lib/torii`. I'm thinking maybe I need to register `test-support` to be made available somehow?

bantic:
Ah... :/  This is due to the fact that torii uses a custom build system (grunt), not ember-cli as its build tool. It should be possible to update the grunt to properly include the test-support tree in the torii tests, and the index.js probably will also need to be informed about the path to the test tree (around [here](https://github.com/Vestorly/torii/blob/master/ember-addon/index.js#L7)). Ideally we can move this whole build system over to ember-cli one day.

jagthedrummer:
Ah-ha! I had looked and grepped through `ember-cli-aptible-shared` trying to figure out where they had registered/exported/whatevered `test-support`. Makes sense now that I couldn't find it. Thanks for the pointer on grunt. I'll see what I can do there. (Realizing that ember-cli shields me from a LOT of stuff.)

jagthedrummer:
OK, I think I finally have the whole `test-support` thing ironed out.

I'm able to import the helper into the `torii` test suite like this:

```javascript
import {stubValidSession} from 'test-support/helpers/torii';
```

And in a host app like this:

```javascript
import {stubValidSession} from 'app-name/tests/helpers/torii';
```

Hopefully that means this is ready to go. Unless we can do something about the `_container_` issue.

README.md
```diff
@@ -435,6 +435,75 @@ Torii will first look for an adapter matching the provider name passed to
 for an adapter at `torii-adapters/geocities`). If there is no matching adapter,
 then the session object will fall back to using the `application` adapter.
 
+## Test Helpers
+
+For testing code that interacts with torii it can be usefull to stub a
```
bantic: spelling nitpick: "useful"

README.md
```diff
@@ -435,6 +435,75 @@ Torii will first look for an adapter matching the provider name passed to
...
+
+with `index.hbs`
+
+```handlebars
+{{#if session.isAuthenticated}}
+  <p>Signed in as : {{session.currentUser.handle}}</p>
+{{else}}
+   <button {{action "signIn" "github"}}>Sign in with github</button>
+{{/if}}
+```
```
bantic: I :clap: the verbose docs. Imo, though, mentioning the `stubValidSession` import and showing a single usage of it is sufficient for this code block, like:
```
import {stubValidSession} from 'your-app/tests/helpers/torii';

// test boilerplate

test('shows something when signed in', function(assert) {
  stubValidSession(application, {object});
  visit('/');

  andThen(function() {
    // make assertions here assuming that the session is set with the object passed above...
  });
});
```

bantic:
@jagthedrummer This is great. I am +1. @mixonic This look good to you or do you have any reservations about using `application.__container__` still?

jagthedrummer:
Ok, just made those couple of changes to the README.

mixonic:
:fireworks: :fireworks: :fireworks: :fireworks: :fireworks:

jagthedrummer:
Woo-hoo!

PR #204, opened 2015-07-22
Added injection of session to component

seawatts:
My use case was that I wanted to be able to access the session from within a component. Currently I had to have this happen in a route or controller. Adding this line makes it so that it is injected into all components. Let me know what you think.

### DIFF
--- a/lib/torii/bootstrap/session.js
+++ b/lib/torii/bootstrap/session.js
@@ -5,6 +5,7 @@ export default function(container, sessionName){
   container.injection('torii:session', 'torii', 'torii:main');
   container.injection('route',      sessionName, 'torii:session');
   container.injection('controller', sessionName, 'torii:session');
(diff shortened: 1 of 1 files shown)


mixonic:
@seawatts I'd prefer to see the Torii session exported as an Ember session so you can explicitly import it:

```js
export default Ember.Component.extend({
  session: Ember.inject.service()
});
```

Can you work up a test and PR for that solution? I agree that component may want access to the session.

seawatts:
@mixonic I agree that should probably be the long term solution. However for the interim do you think we could just do this?

mixonic:
@seawatts It would be painful to add support for the session everywhere only to remove it in a later release. I'd strongly prefer to see the `inject.service` solution.

You should be able to do the injection in your own app if you want a quick solution without a solution upstream.

mixonic:
Addressed by https://github.com/Vestorly/torii/pull/214

PR #206, opened 2015-07-27
For Discussion : Iframe Service

jagthedrummer:
I got a very basic iframe implementation working and I wanted to get some feedback before I get too far down the rabbit hole. At this point I just duplicated the Popup service and made the minimum changes required to get an iframe approach to work in a test app. Issues that could use discussion are:

*  There's a lot of code that I didn't touch, which I'm thinking should probably be extracted into a mix-in that both services can use.

*  I'd love to hear some thoughts about how the iframe service should be used instead of the popup service. For testing this implementation I just registered the iframe service AS the popup service in `lib/torii/bootstrap/torri.js`. Maybe we just add documentation about how someone would override this in an application specific initializer? Another route would be to register it as the iframe service and then create a new batch of providers that know to look for the iframe service instead of the popup service, but that seems like a bunch of unnecessary code. To my mind a provider shouldn't care whether the OAuth flow is happening via popup or iframe. That seems more like an application configuration option.

* Managing the iframe needs work. At this point when you `open` the iframe service I'm just unceremoniously appending an `iframe` element to the end of the `body`, with an `id` of `torii-iframe`. Then when you `close` the service I'm just removing it. This seems.... inelegant. Anybody have any thoughts about how we should approach this? Maybe you pass an element or element id to `open` so that we know where to put the iframe? Maybe we require you (or give you the option) to transition to some route where you setup the iframe yourself? Also, with the popup service the user could close the window to 'opt out' of signing in. Do we need to offer something similar for the iframe?


jagthedrummer:
I just realized that we'll also need a way to make the `RedirectHandler` not call `windowObject.close()`. For some reason it's not happening to me, but a user reported that after the iframe returned from the provider that their tab was automatically closed. After looking at it I can totally see why, but I'm not sure why it's not happening in any of my browsers.

jagthedrummer:
I went ahead and did some refactoring to eliminate all the duplicate code. I think this makes it much easier to actually see what's going on. I pulled most of the code from `lib/torii/services/popup.js` into `lib/torii/mixins/services-mixin.js`. Now `popup.js` contains only the code that's directly related to opening and closing the popup window itself. Likewise `iframe.js` only deals with the iframe and not any of the other setup. The `services-mixin.js` contains the main `open` method, which still does all of the setup for remote URL options, setting the request ID and what not. Then instead of opening a popup/iframe directly it calls `openImpl` which is defined in each service. Each service implements `close` to handle tearing down the popup or iframe.

lib/torii/mixins/services-mixin.js
```diff
@@ -0,0 +1,161 @@
+import UUIDGenerator from 'torii/lib/uuid-generator';
```
mixonic: Could we name this file `ui-service-mixin.js` or something else more specific than `services-mixin`?

lib/torii/mixins/services-mixin.js
```diff
@@ -0,0 +1,161 @@
...
+      }
+      optionsStrings.push(
+        key+"="+value
+      );
+    }
+  }
+  return optionsStrings.join(',');
+}
+
+function prepareOptions(options){
```
mixonic: Seems like this should not be shared

lib/torii/mixins/services-mixin.js
```diff
@@ -0,0 +1,161 @@
...
+      localStorage.setItem(CURRENT_REQUEST_KEY, pendingRequestKey);
+
+
+      service.openImpl(url, pendingRequestKey, optionsString);
+
+      if (service.remote && !service.remote.closed) {
+        service.remote.focus();
+      } else {
+        reject(new Error(
+          'iframe could not open or was closed'));
```
mixonic: This error does not seem shareable either.

Perhaps `open` should be split into several functions (perhaps each returns a promise perhaps not) that could make overriding or changing the calls simpler.

lib/torii/mixins/services-mixin.js
```diff
@@ -0,0 +1,161 @@
...
+  pollRemote: function(){
+    if (!this.remote) {
+      return;
+    }
+    if (this.remote.closed) {
+      this.trigger('didClose');
+    }
+  },
+
+  schedulePolling: function(){
```
mixonic: Would also not be needed for iframes.

lib/torii/bootstrap/torii.js
```diff
@@ -27,7 +28,7 @@ export default function(container){
   container.register('torii-provider:edmodo-connect', EdmodoConnectProvider);
   container.register('torii-adapter:application', ApplicationAdapter);
 
-  container.register('torii-service:popup', PopupService);
+  container.register('torii-service:popup', IframeService);
```
mixonic: I suggest registering `torii-service:iframe` as another service, regardless of how we structure the APIs for the end user.

lib/torii/services/iframe.js
```diff
@@ -0,0 +1,23 @@
+import ServicesMixin from 'torii/mixins/services-mixin';
+
+var on = Ember.on;
+
+var Iframe = Ember.Object.extend(Ember.Evented, ServicesMixin, {
+
+  // Open an iframe window.
+  openImpl : function(url, pendingRequestKey, optionsString){
```
mixonic: What is `Impl`? Can we avoid shortening words if this is supposed to be `implementation`?

test/tests/unit/services/iframe-test.js
```diff
@@ -0,0 +1,109 @@
...
+  setTimeout(function(){
+    ok(!promise.isFulfilled, 'promise is not fulfulled by invalid data');
+    deepEqual("http://authServer",
+        localStorage.getItem(PopupIdSerializer.serialize("invalid")),
+        "doesn't remove the key from local storage");
+    start();
+  },10);
+});
+
+//asyncTest("open rejects when window closes", function(){
```
mixonic: I'm unsure what the lifecycle is of the iframe, this may or may not be needed.

mixonic:
The config or developer experience using this is still hazy. Appending the iframe to the body is hardly ideal. Perhaps `open` could be passed the parent DOM node of the iframe as an argument and the iframe service would append to that. Need to ponder some more.

lib/torii/mixins/services-mixin.js
jagthedrummer: :+1:

lib/torii/mixins/services-mixin.js
jagthedrummer: Ah, dang! I had meant to set it up so that the actual error thrown would pull the message from an attribute, which could be overridden in each service, but I forgot to. Would that work?

lib/torii/services/iframe.js
jagthedrummer: Yep, no problem.

test/tests/unit/services/iframe-test.js
jagthedrummer: I'm a little unsure on that myself.

lib/torii/mixins/services-mixin.js
mixonic: I think there should be a way to structure these specifics with composition instead of overriding properties and hooks. `open` should call a number of more specific functions, then each implementation could customize what is called and how.

jagthedrummer:
Yeah, adding the iframe to the body is not great. I think passing in the parent node makes sense. And maybe we fall back to `body` if you don't pass anything.

lib/torii/mixins/services-mixin.js
jagthedrummer: Could you elaborate on this?

lib/torii/bootstrap/torii.js
jagthedrummer: What do you think about this? We register both services under their appropriate name, and then based on a config option we register `torii-service:remote` to be one or the other. Then we alter the providers to look for the `remote` service instead of the `popup` service.

```javascript
  container.register('torii-service:iframe', IframeService);
  container.register('torii-service:popup', PopupService);

  var remoteServiceName = configuration.remoteServiceName || 'popup';
  if(remoteServiceName === 'iframe'){
    container.register('torii-service:remote', IframeService);
  }else{
    container.register('torii-service:remote', PopupService);
  }
```

lib/torii/mixins/services-mixin.js
jagthedrummer: Ah, never mind, I see what you mean. This is for preparing the list of options for `window.open` which we don't need in the iframe version.

lib/torii/mixins/services-mixin.js
jagthedrummer: OK, I'll take a stab at that.

lib/torii/mixins/services-mixin.js
jagthedrummer: I just discovered that the popup service still works even when `schedulePolling` is never called. Does `schedulePolling` just cover an edge case of some sort?

jagthedrummer: Derp. It's for the case where the popup is closed, not the happy path.

jagthedrummer:
So, thinking through the iframe use case some more, if we have it so that you have to pass in a DOM node, I'm imagining that the flow will go something like this:

* User click 'Sign in' firing some `signIn` action
* The `signIn` action opens an in-window modal of some sort
* The `signIn` action calls `open` with something like `this.get("session").open("clickfunnels-oauth2","#modal-inner-content");`
* We append the iframe to `#modal-inner-content`

I'm further imagining that the modal will have some sort of UI element to allow a user to decide not to sign in (some way to close the modal). In that case we'd either need to poll the iframe to see if it has disappeared, much like we do with the popup, or we'd need to instruct people to manually trigger `didClose` in the method that's responsible for tearing down the modal.

Thoughts on the best approach?

jagthedrummer:
With the latest changes in this branch and the code below I've been able to put together a reasonable user experience for the iframe flow. When a user clicks 'Sign in' a modal opens with a login form. If the user signs in successfully the modal goes away and they're logged in. If the user clicks outside the modal form, the modal goes away and the torii promise rejects. 

With this implementation, when the modal is closed it's up to the application developer to remove the `#torii-iframe` element from the DOM so that the `pollRemote` method will see that it's gone. This works, but it doesn't feel quite right.

I'm sure it can use some more polish under the hood. I'd love to hear feedback.

```javascript
// app/routes/application.js
import Ember from 'ember';

export default Ember.Route.extend({

  beforeModel : function(){
    return this.get("session").fetch().catch(function() { console.log('in the fetch callback catch!'); });
  },

  actions: {
    signIn: function() {
      Ember.$('body').addClass('signingIn');
      Ember.$('#signin-modal-back').one('click',function(){
        Ember.$('#signin-modal-content').html('');
        Ember.$('body').removeClass('signingIn');
      });
      this.get("session")
        .open("clickfunnels-oauth2",{ iframeParent : '#signin-modal-content'})
        .then(function(){
          Ember.$('body').removeClass('signingIn');
        });
    },

    logout: function() {
      this.get("session").close();
    }
  }

});
```

```handlebars
<!-- app/templates/application.hbs -->
{{#if session.isAuthenticated}}
  <a {{action "logout"}} href='#'>Logout</a>
  {{ log session.isAuthenticated }}
  <br/>
  {{outlet}}
{{else}}
  <button {{action "signIn"}}>Sign in</button>
{{/if}}

<div id="signin-modal-back">
  <div id="signin-modal-frame">
    <div id="signin-modal-content">
    </div>
  </div>
</div>
```

And then in `config/environment.js` the `torii` object looks like this:

```javascript
    torii: {
      remoteServiceName: 'iframe',
      providers : {
        'acme-oauth' : {
          apiKey : 'xxx'
        }
      }
    }
```

And just for good measure, the CSS for the modal:

```css
iframe{
  width:100%;
  height:500px;
  border:1px solid blue;
}

#signin-modal-back{
  display:none;
}

body.signingIn #signin-modal-back{
  display:block;
  position:absolute;
  left:0;
  top:0;
  width:100%;
  height:100%;
  background:black;
  z-index:1000;
}

body.signingIn #signin-modal-frame{
  position:absolute;
  top:100px;
  left:50%;
  margin-left:-250px;
  width:500px;
  height:500px;
  padding:10px;
  background:#dedede;
}
```

jagthedrummer:
I've been playing with this some more and have an alternate iframe implementation working with a liquid fire modal.

Along with the changes mentioned above to `config/environment.js` I have:

In `router.js`
```javascript
this.modal('signin-modal', { withParams: 'signIn' });
```

In `controllers/application.js`
```javascript
queryParams: ['signIn'],
signIn: null
```

In `routes/appication.js`
```javascript
  beforeModel : function(){
    return this.get("session").fetch().catch(function() { console.log('in the fetch callback catch!'); });
  },

  actions: {
    signIn: function() {
      var _this = this;
      this.controller.set('signIn','');
      Ember.run.next(this,function(){
        this.get("session")
          .open("clickfunnels-oauth2",{ iframeParent : '#signin-modal-content'})
          .then(function(){
            console.log('we are signed in....');
            _this.controller.set('signIn',null);
          },function(error){
            console.log('sad panda', error);
          });
       });
    },

    logout: function() {
      this.get("session").close();
    }
  }
```

And in `templates/components/signin-modal.hbs`
```handlebars
<div id="signin-modal-content">
</div>
```

mixonic:
@jagthedrummer I will have time to give this some attention this week. I just want to make sure the API design is something we can live with for a while.

jagthedrummer:
Great! Thanks for the update. So far this has been working well for us.

jagthedrummer:
I just ran into a situation where we're going to want to be able to opt in to the iframe service on a per provider basis. For our own "internal login" we want to use the iframe, but we also need to provide for Stripe Connect. Stripe has `X-Frame-Options` set to `SAMEORIGIN`, so an iframe won't work for them. I'll be looking at what it will take to set this at the provider level instead of for the whole app.

jagthedrummer:
Hey guys, just wondering if there's anything I can do to advance this. The `master` branch has changed quite a bit since I submitted this PR and I'm currently in the process of getting this branch up to date with all the latest stuff. I'd sure love to avoid having to do that more that once if possible.

jagthedrummer:
OK, I have a branch of this that's up to date with the current `master` branch. Should I force push over this branch/PR or push a new branch and open a new PR?

We ended up handling Stripe in a different way, so I didn't do anything for making it possible to configure popup vs iframe on a provider by provider basis. I suspect that it would be needed in some cases, and I'd be happy to help figure out how to make that happen. In the interest of not needing every single provider to check the config to know which 'popup' to request I'm thinking that during initialization we'd need to override the 'popup' object in any providers that are configured to use the iframe remote. If that's a reasonable approach, it seems like `lib/torii/instance-initializers/walk-providers` would be the place to do it. Does that sound right?

jagthedrummer:
Sounds good. Closing this PR and will open a new one shortly.

jagthedrummer:
Also, some time on Friday morning should work for me.

PR #214, opened 2015-08-15
Updates Torii singletons to services

pzuraq:
Still needs:

- [x] Test coverage of `inject.service()`
- [ ] Updates to version numbers and documentation

Opening this PR for discussion. This PR aims to bring the various Torii singletons inline with Ember Services by using the `Ember.Service` class and registering the singletons using the standard `service:*` naming scheme.

The biggest drawback to this change is breaking changes to anyone currently using the Torii singletons directly (i.e. through `container.lookup`). However, this functionality has been deprecated and the `inject` pattern has been introduced to fill that gap.

Alternatively, we could register the service under both names and maintain backwards compatibility (if this is possible).

### DIFF
--- a/lib/torii/bootstrap/session.js
+++ b/lib/torii/bootstrap/session.js
@@ -1,10 +1,11 @@
-import Session from 'torii/session';
+import ToriiSessionService from 'torii/services/torii-session';
 
 export default function(container, sessionName){
-  container.register('torii:session', Session);
-  container.injection('torii:session', 'torii', 'torii:main');
-  container.injection('route',      sessionName, 'torii:session');
-  container.injection('controller', sessionName, 'torii:session');
+  var sessionFactoryName = 'service:' + sessionName;
+  container.register(sessionFactoryName, ToriiSessionService);
+  container.injection(sessionFactoryName, 'torii', 'service:torii');
+  container.injection('route',      sessionName, sessionFactoryName);
+  container.injection('controller', sessionName, sessionFactoryName);
 
   return container;
 }
--- a/lib/torii/bootstrap/torii.js
+++ b/lib/torii/bootstrap/torii.js
@@ -1,4 +1,3 @@
-import Torii from 'torii/torii';
 import LinkedInOauth2Provider from 'torii/providers/linked-in-oauth2';
 import GoogleOauth2Provider from 'torii/providers/google-oauth2';
 import GoogleOauth2BearerProvider from 'torii/providers/google-oauth2-bearer';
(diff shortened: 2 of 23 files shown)


lib/torii/bootstrap/session.js
```diff
@@ -1,10 +1,10 @@
-import Session from 'torii/session';
+import ToriiSessionService from 'torii/services/torii-session';
 
 export default function(container, sessionName){
-  container.register('torii:session', Session);
```
mixonic: Can you use `sessionName` for the name? We don't want to leak `torii` into the app code for the session service.

```
let sessionFactoryName = `service:${sessionName}`;
container.register(sessionFactoryName, ToriiSessionService);
container.injection(sessionFactoryName, 'torii', 'service:torii');
container.injection('route', sessionName, sessionFactoryName);
container.injection('controller', sessionName, sessionFactoryName);
```

lib/torii/bootstrap/torii.js
```diff
@@ -11,10 +10,12 @@ import AzureAdOauth2Provider from 'torii/providers/azure-ad-oauth2';
 import StripeConnectProvider from 'torii/providers/stripe-connect';
 import EdmodoConnectProvider from 'torii/providers/edmodo-connect';
 
+import ToriiService from 'torii/services/torii';
 import PopupService from 'torii/services/popup';
 
-export default function(container){
-  container.register('torii:main', Torii);
+export default function(container) {
+  container.register('service:torii', ToriiService);
```
mixonic: Ah interesting.

lib/torii/initializers/initialize-torii-session.js
```diff
@@ -8,7 +8,7 @@ export default {
   initialize: function(container){
     if (configuration.sessionServiceName) {
       bootstrapSession(container, configuration.sessionServiceName);
-      container.injection('adapter', configuration.sessionServiceName, 'torii:session');
+      container.injection('adapter', configuration.sessionServiceName, 'service:torii-session');
```
mixonic: Again this would be the `sessionFactoryName` based on `configuration.sessionServiceName`

lib/torii/services/torii-session.js
```diff
@@ -11,7 +11,7 @@ function lookupAdapter(container, authenticationType){
   return adapter;
 }
 
-var Session = Ember.ObjectProxy.extend({
+export default Ember.Service.extend(Ember._ProxyMixin, {
```
mixonic: :+1: Someday I will need to stop using that proxy mixin.

test-support/helpers/torii.js
```diff
@@ -1,5 +1,5 @@
 export function stubValidSession(application,currentUser) {
-  var session = application.__container__.lookup('torii:session');
+  var session = application.__container__.lookup('service:torii-session');
```
mixonic: `sessionFactoryName`

mixonic: Well, in tests I would be fine with hard-coding it.

mixonic:
I am fine with changing `torii:main` to `service:torii`, since `torii` is still the injection on routes.

I'd like the configuration for service name to apply to the name in the container. Really, the service should
be opt-in by configuring `sessionServiceName`, and then available as that same name for injection.

Basically this allows:

```js
/* jshint node: true */
// config/environment.js
module.exports = function(environment) {
  var ENV = {
    /* ... */
    torii: {
      // a 'session' property will be injected on routes and controllers, and
      // a service by this name be made available
      providers: { /* ... */ }
    }
  };
  return ENV;
};
```

```js
export Ember.Component.extend({
  session: Ember.inject.service()
});
```

And `torii-service` or something like it doesn't need to leak into component code.

This is great @pzuraq!

pzuraq:
@mixonic updated, and tests added! Let me know if anything else needs to be changed.

mixonic:
Thank you @pzuraq! Great work on this! I will need to add a `CHANGELOG` to call out the breaking change in the next release.

PR #216, opened 2015-08-18
add openid-connect with jwt tokens

albertjan:
Adds a generic openid connect provider.

 - decodes the jwt to a javascript object
 - doesn't validate it that would mean adding a __lot__ of dependencies for certificate validation
 - tested irl with auth0 also borrowed some jwt decode tests from them

There are three places I've sinned:

 -


albertjan:
I don't think I caused the test to fail ... they don't fail on my machine and I didn't touch the initializer.

mixonic:
@albertjan I kicked the build.

lib/torii/bootstrap/torii.js
```diff
@@ -10,6 +10,7 @@ import GithubOauth2Provider from 'torii/providers/github-oauth2';
 import AzureAdOauth2Provider from 'torii/providers/azure-ad-oauth2';
 import StripeConnectProvider from 'torii/providers/stripe-connect';
 import EdmodoConnectProvider from 'torii/providers/edmodo-connect';
+import OpenidConnect from 'torii/providers/openid-connect';
```
mixonic: Nitpick, but `OpenIDProvider` seems fine. "connect" is a specific thing for stripe, you don't need to use it here.

lib/torii/lib/uuid-generator.js
```diff
@@ -8,8 +8,6 @@ var UUIDGenerator = {
     });
     return uuid;
```
mixonic: If you drop this whitespace plz drop this `,` as well. Seems weird it got by jshint

mixonic:
@albertjan How coupled to JWT is OpenID? I don't think you need JWT in order to do OpenID.

And can you link to what asserting the certificate might look like? If this provider skips any important security requirements of OpenID, I'm not sure we would want to include it.

mixonic:
If you're targeted at auth0 and they are the ones that use JWT, you could also write a generic OpenID provider and an auth0 provider that subclasses that.

mixonic:
Regarding the issue with setting `optionalUrlParams` to nothing- perhaps this should not be subclassing the `Oauth2` provider, since it really is not OAuth2.

albertjan:
I was attemptping to implement the "OpenID Connect" flow as described here. Which is as far as I know tightly coupled with jwt. See: http://openid.net/specs/openid-connect-core-1_0.html#IDToken. 

I don't know if there is such a thing as a super class OpenID of which OpenID Connect is an derivative.

About the certificate validation:

 - In the current state you can't assert the senders identify
 - We currently validate the token on the server

OpenID Connect is based on part of the OAuth spec.

mixonic:
My brain just exploded. Thank you for the context and link. What a mess of naming schemes.

albertjan:
Hahah yeah, these guys keep on going :smile:.

albertjan:
We'd need to include this lib: https://kjur.github.io/jsrsasign/ if we'd want to add javascript world asymmetric signature verification.

albertjan:
@mixonic How do you want me to proceed? :smile:

mixonic:
@albertjan is that library even helpful? Wouldn't confirming the signature mean providing the secret to JavaScript?

I think of this like OAuth2 authorization code flows- The code is just provided to your session abstraction, or if you write a custom provider you can confirm the code in a provider. But we don't attempt to do what is by design a server-side step of the auth on the client.

If this is true (that we would leak the private key), then I'd prefer to just return the JWT as data, unparsed. If someone wants to opt into to trusting a JWT implicitly (without confirmation it is authentic) they can still do that, but I'm not comfortable prescribing it by doing the parsing for them.

albertjan:
@mixonic if you use asymmetric key encryption you can verify the signature of the token with the public key. That s what you'd use that library for.

> I'd prefer to just return the JWT as data, unparsed. If someone wants to opt into to trusting a JWT implicitly (without confirmation it is authentic) they can still do that, but I'm not comfortable prescribing it by doing the parsing for them. 

That sounds reasonable, and a good idea! :smile:

I'll go down that path and I could write a piece for the docs where I explain how you could use it with jsrasign to verify it. And how you'd decode the token.

mixonic:
@albertjan that sounds great- docs to highlight the path forward. :+1: 

Thanks much for your patience as I got around to this. Excited to see it complete :-)

albertjan:
I added some docs, we'll probably need to go over it a few times as I'm not a native speaker. So don't hold back. :smile:

albertjan:
I also figured out why the build failed. The bower packages have references to each other that don't line up.

albertjan:
```sh
$ grunt test
Running "clean:build" (clean) task
>> 1 path cleaned.

Running "transpile:amd" (transpile) task

Running "concat:forTests" (concat) task
File tmp/torii.testbuild.js created.

Running "jshint:lib" (jshint) task
>> 44 files lint free.

Running "transpile:testsAmd" (transpile) task

Running "concat:amdTests" (concat) task
File tmp/tests.amd.js created.

Running "testem:ci:basic" (testem:ci) task

  .....................................................................................................................................................

  149 tests complete (3945 ms)


Done, without errors.
```

I really don't get why I can't seem to get the tests to pass at circleci. And I also don't understand why I have one test less. weird

albertjan:
Hey @mixonic :smile: you haven't forgotten about this have you?

albertjan:
I seem to have an ongoing fight with the tests. Do you have any idea what I'm doing wrong?

bantic:
@albertjan I'd like to help get an OpenID provider into Torii, it would be great. Can you explain a little more about the flow? Where does the `pem` that you mention in the Readme come from? Is that provided by the OpenID Provider when you create an application (at say, google)?

albertjan:
@bantic Hi yes officially a openid provider would have a .well-known enpoint where you could get the publickey/certifiicate with which the jwt is signed. Most of the time however they don't have cors headers on those endpoints. Making it impossible to get the publickey/certificate in a web app.

The flow is as follows:

1. user goes to secure route
2. torii opens a popup to the auth providers loginscreen
3. that replies with a query param `?id_token` which contains the jwt
4. this result is used to authenticate the user to the api
5. it is then an optional step for the ember app to do a verification of the jwt itself.

the last step I had originally included in the PR but that added a couple of dependencies and since it's optional we decided it was better to document then to build.

Step 5 could also be replaced with a roundtrip to the api, to verify the signature there.

albertjan:
@mixonic @bantic Do you want me to rebase this? Does it have any chance of making it into torii?

mixonic:
@albertjan Howdy. I focused on a few other backlog items the last few releases- I believe @bantic was reviewing this in the fall and I'll confer about it with him this week.

albertjan:
Awesome, let me know where I can help.

dmitrizagidulin:
How's this PR doing? I'm in the process of switching our Authn stack to OpenID Connect, and would love to have this Torii plugin for Ember apps. Would be happy to collaborate / add an extra pair of eyes on this PR.

jsmestad:
Anything we can do to help move an OIDC provider along?

dmitrizagidulin:
@jsmestad I could possibly give it a shot. What's your use case?

albertjan:
I think everything is here it needs a rebase and probably a bit of updating to remove some deprecations, but knowing ember it'll all still work. 

I'm quite sure the guys @curit still use it to authenticate against azure. 

The only big thing I know changed is that the queryparam response code is deprecated in favour of a url fragment. So you'll have to work around that in torii. 

Its a shame this didn't fet merged but I understand too @vestorly or @201created probably don't use oid and it would therefor have been a burden. I never fully understood why the tests didn't pass on the ci server. 🤷‍♂️

PR #217, opened 2015-08-23
Add CSRF protection for OAuth2 providers via state parameter

alisdair:
### Summary

The OAuth2 code provider (and all subclasses) now default to a generated random "state" URL parameter. This is transmitted to the server, and the provider validates that the same state parameter is received in the response.

This protects against Cross-Site Request Forgery. For more details on this part of the OAuth2 spec, see here RFC 6749 section 10.12:

http://tools.ietf.org/html/rfc6749#section-10.12

### Implementation

The OAuth2 code provider has a default value for `state` which generates and caches a random URL-safe string. This can be overridden by the configuration if the application requires it.

There's also a new mock popup test helper, which defaults to returning the `state` parameter from the request as part of the resolved `authData`. All relevant integration tests have been updated to use this new helper.

I also moved `buildUrl`-related integration tests for the Google and Stripe providers into unit tests, where they are easier to assert.

### Open issues

I haven't tried checking that the providers work against the real APIs, but I checked all the developer docs for the providers I updated and confirmed that they do support the state parameter as per spec.

Also not sure how to update the documentation for this change.

### DIFF
--- a/lib/torii/lib/random-url-safe.js
+++ b/lib/torii/lib/random-url-safe.js
@@ -1,0 +1,11 @@
+export default function randomUrlSafe(length) {
+  var chars = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+  var result = '';
+
+  for (var i = 0; i < length; i++) {
+    result += chars.charAt(Math.floor(Math.random() * chars.length));
+  }
+
+  return result;
+}
+
--- a/lib/torii/providers/azure-ad-oauth2.js
+++ b/lib/torii/providers/azure-ad-oauth2.js
@@ -25,7 +25,7 @@ var AzureAdOauth2 = Oauth2.extend({
   responseMode: configurable('responseMode', null),
 
   responseParams: computed(function () {
-    return [ this.get('responseType') ];
+    return [ this.get('responseType'), 'state' ];
   }),
 
   apiVersion: '1.0',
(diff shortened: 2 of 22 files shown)


mixonic:
@alisdair checking the demo locally now.

mixonic:
Fails to work with Facebook Oauth2, LinkedIn Oauth2, Google Oauth2, GitHub Oauth2. Something seems awry.

alisdair:
Thanks for checking, and oops, sorry :disappointed:. If it's not immediately obvious what's gone wrong, I can take a look myself later in the week.

mixonic:
Yeah I will try to find some time ASAP

alisdair:
I think it's fixed now. I was only able to test GitHub and Google, but the same error was in all providers—I forgot to add the `state` parameter to `responseParams`, so the popup wasn't resolving it back to the provider.

I updated the oauth2-code provider so that it only validates the state param if it's in the list of response params. Also updated the mock popup so that it only sends back state if it's in the list.

Tests pass, the demo seems to work, and the tests do also cover the change I just made.

Still haven't had time to test this in my own app, will do that when I next get some time.

alisdair:
@mixonic Confirmed this branch is working for my app. [This PR is in production with my branch of torii](https://github.com/alisdair/team-time-zone/pull/2), and [you can verify it working yourself here](https://teamtime.zone).

mixonic:
@alisdair thank you for the continued attention to this. I will give a quick review this afternoon, hopefully merge and release :rocket:

mixonic:
Thank you @alisdair! Very valuable and well done contribution. I appreciate all your work :-)

PR #224, opened 2015-09-08
Allow unregistered factorys to require authentication

mixonic:
Routing options should not require that the factory actually be present in the container when bootstrapping.

### DIFF
--- a/lib/torii/bootstrap/routing.js
+++ b/lib/torii/bootstrap/routing.js
@@ -1,17 +1,25 @@
 import ApplicationRouteMixin from 'torii/routing/application-route-mixin';
 import AuthenticatedRouteMixin from 'torii/routing/authenticated-route-mixin';
 
+var AuthenticatedRoute = null;
+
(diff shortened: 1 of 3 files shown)


lib/torii/bootstrap/routing.js
```diff
@@ -1,17 +1,25 @@
...
 
-  var ApplicationRoute = container.lookup('route:application');
-  ApplicationRoute.reopen(ApplicationRouteMixin);
+function reopenOrRegister(container, factoryName, mixin) {
+  var factory = container.lookup(factoryName);
+  if (factory) {
+    factory.reopen(mixin);
+  } else {
+    if (!AuthenticatedRoute) {
+      AuthenticatedRoute = Ember.Route.extend(AuthenticatedRouteMixin);
```
rwjblue: This should use `container.lookup('route:basic')` (which is what Ember itself uses for generated routes).

rwjblue: https://github.com/Vestorly/torii/pull/226

PR #226, opened 2015-09-09
Honor `route:basic` convention for generated routes.

rwjblue:
By default Ember uses `route:basic` if a route for the specific routename is not found. This emulates that behavior...


mixonic:
@rwjblue will `basic` always be present?

rwjblue:
Yes, it is injected into the default registry/container created.

PR #228, opened 2015-09-10
Added fix for dynamic sessionServiceName

bmeyers22:


lib/torii/routing/application-route-mixin.js
```diff
@@ -11,7 +13,7 @@ export default Ember.Mixin.create({
     }
   },
   checkLogin: function () {
-    return this.session.fetch()
+    return this[configuration.sessionServiceName].fetch()
```
mixonic: I'd prefer `this.get(configuration.sessionServiceName).fetch()` - just a little safer.

lib/torii/routing/authenticated-route-mixin.js
```diff
@@ -11,11 +13,12 @@ export default Ember.Mixin.create({
     }
   },
   authenticate: function (transition) {
-    var route = this;
-    var isAuthenticated = this.get('session.isAuthenticated');
+    var route = this,
+      session = this[configuration.sessionServiceName],
```
mixonic: Again `this.get(configuration.sessionServiceName)`

lib/torii/routing/authenticated-route-mixin.js
```diff
@@ -11,11 +13,12 @@ export default Ember.Mixin.create({
...
+      isAuthenticated = this.get(configuration.sessionServiceName + '.isAuthenticated');
```
mixonic: And then this can `session.get('isAuthenticated')`

mixonic:
:+1: good catch @bmeyers22. I would love a test if you could add one (should be fairly easy to add).

mwpastore:
@mixonic Won't the tests for the `authenticatedRoute()` DSL cover these changes?

bmeyers22:
As far as tests, I believe the correct place would be the configuration tests, which should already cover that configs defined can be read properly from the module

mixonic:
Seems like a basic test in https://github.com/Vestorly/torii/blob/master/test/tests/acceptance/routing-test.js would be appropriate, but with the session service name (like [here](https://github.com/Vestorly/torii/blob/master/test/tests/acceptance/routing-test.js#L21)) set to something other that `session`

bmeyers22:
@mixonic I wrote a test, let me know if this was the kind of thing you were looking for

mixonic:
@bmeyers22 perfect! Thank you. Can you rebase down to single squashed commit? I'd love to merge, I will push out beta.5 today and hopefully that is the last.

bmeyers22:
All set

mixonic:
:heart_eyes: thank you for the contribution!

mixonic:
In 0.6.0-beta.7

PR #234, opened 2015-09-18
Only use has if it exists on the application

bmeyers22:
Fix for if using a version of ember without the has function on the application

### DIFF
--- a/lib/torii/bootstrap/torii.js
+++ b/lib/torii/bootstrap/torii.js
@@ -34,7 +34,9 @@ export default function(application) {
 
   if (window.DS) {
     var storeFactoryName = 'store:main';
-    if (application.has('service:store')) {
+    if (application.has && application.has('service:store')) {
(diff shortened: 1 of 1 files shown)


lib/torii/bootstrap/torii.js
```diff
@@ -34,7 +34,9 @@ export default function(application) {
 
   if (window.DS) {
     var storeFactoryName = 'store:main';
-    if (application.has('service:store')) {
+    if (application.has && application.has('service:store')) {
+      storeFactoryName = 'service:store';
+    } else if (application.registry.registrations['service:store']) {
```
marcoow: There's also `application.hasRegistration` in 2.1 (see http://emberjs.com/blog/2015/08/16/ember-2-1-beta-released.html#toc_registry-and-container-reform), I guess that should be used if available.

mixonic:
Please check 0.6.0-beta.7 :-) Thank you!

PR #237, opened 2015-09-22
Pass _initRouterJs arguments to _super.

xtian:
The incorrect `_super` call here breaks loading substate code that depends on the `moduleBasedResolver` flag: http://git.io/vnEeC

### DIFF
--- a/lib/torii/router-dsl-ext.js
+++ b/lib/torii/router-dsl-ext.js
@@ -11,7 +11,7 @@ proto.authenticatedRoute = function() {
 Router.reopen({
   _initRouterJs: function() {
     currentMap = [];
-    this._super();
+    this._super.apply(this, arguments);
     this.router.authenticatedRoutes = currentMap;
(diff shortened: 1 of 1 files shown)


lib/torii/router-dsl-ext.js
```diff
@@ -11,7 +11,7 @@ proto.authenticatedRoute = function() {
 Router.reopen({
   _initRouterJs: function() {
     currentMap = [];
-    this._super();
+    this._super.apply(this, arguments);
```
erikreedstrom: Can we tag with something to remove when torii is fixed?

erikreedstrom:
:+1:

mixonic:
Definitely.

mixonic:
Thank you!

mixonic:
Fixed in 0.6.0-beta.8

xtian:
Thanks!

PR #240, opened 2015-09-28
Ensure that `session.attemptedTransition` is set

bantic:
The `session.attemptedTransition` property is currently set by the authenticated route mixin in its
`authenticate` method when `isAuthenticated === undefined`. `isAuthenticated` will only be `undefined` before the first attempt to `fetch` the session by the adapter. After that initial attempt, `isAuthenticated` will either be boolean `false` or `true`.

This changes the setting of `attemptedTransition` to happen regardless
of whether we have previously attempted to fetch authentication or not,
ensuring that upstream routes will be able to `attemptedTransition.retry()` to return to the route the user had been attempting to visit while unauthenticated.


mixonic:
Ah, great catch! Thanks @bantic

PR #244, opened 2015-10-12
convert to ember-cli

bantic:
This converts Torii to use ember-cli as its build system.
Most of the change is fairly mechanical: moving files to be in the places ember-cli expects them to be. The initializers and services move into `app/` and most of the other files move to `addon/`.

Some of the test setup is different in an ember-cli environment, though. The most impactful change is to the way configuration is handled. Previously, configuration was read off a window.ENV.torii global. Now it is read from the `config/environment` export that ember-cli provides. The `torii/configuration.js` file is changed to now export `getConfiguration` and `configure`, in addition to `configurable` which it already exported. Its default export is no longer used (it previously default-exported the torii configuration).
The main "torii" initializer uses `configure` to set the torii configuration from `config/environment`.
Many of the tests have been changed to use `configure` rather than modifying a configuration global directly.

Since this is a large changeset, I attempted to minimize some of the more minor changes by modifying the jshintrc files to add some globals (`window`, `screen`, `localStorage`) as well as to set `"unused": false`. In a later PR I intend to revert those settings and fix the associated jshint failures that crop up. I probably should have done more with the test jshintrc also; there's a little bit of a noise in the test changes due to additions of `assert`.

To do:
  * [ ] update this or make a separate PR to remove the old grunt build system
  * [ ] real-world testing
  * [x] use ember-try to test different ember versions
  * [ ] update documentation to point out that globals-mode Torii will no longer be supported when this is an ember-cli addon


addon/configuration.js
```diff
@@ -1,16 +1,18 @@
...
   return Ember.computed(function configurableComputed(){
     // Trigger super wrapping in Ember 2.1.
     // See: https://github.com/emberjs/ember.js/pull/12359
     this._super = this._super || (function(){ throw new Error('should always have _super'); })();
-    var namespace = this.get('configNamespace'),
-        fullKey   = namespace ? [namespace, configKey].join('.') : configKey,
-        value     = get(configuration, fullKey);
+    var configNamespace = NAMESPACE + '.' + this.get('name');
+    var propertyPath = configNamespace + '.' + configKey;
+    let configuration = getConfiguration();
```
mixonic: Haha, harmless but the `var`/`let` interspersing is kind of whack :-)

addon/configuration.js
```diff
@@ -19,13 +21,21 @@ function configurable(configKey, defaultValue){
...
   });
 }
 
-export {configurable};
+function configure(settings) {
+  configuration = settings;
+}
+
+function getConfiguration() {
+  return configuration;
```
mixonic: Why have this function instead of just using `configuration` directly?

Oh, so it can be exported (and I presume used in tests).

addon/providers/base.js
```diff
@@ -1,6 +1,8 @@
 import requiredProperty from 'torii/lib/required-property';
+import { getConfiguration } from 'torii/configuration';
 
 var computed = Ember.computed;
+var get      = Ember.get;
```
mixonic:
```js
let { computed, get } = Ember;
```

meh. :bike: :house:

index.js
```diff
@@ -0,0 +1,16 @@
+/* jshint node: true */
+'use strict';
+
+module.exports = {
+  name: 'torii',
+  isDevelopingAddon: function() {
+    return true;
```
mixonic: We should find out what to do about this. I don't think we should be shipping with it `true`.

We had something similar at work yesterday. Perhaps we can open an issue and/or bug the ember-cli team.

mixonic:
You are a saint. :pray:

addon/configuration.js
bantic: agreed, it looks odd. Part of me trying to minimize diff noise was not changing existing `var` declarations. In this case everything else about the line ended up changing so it was moot. But after a merge I intend to do a cleanup PR.

addon/configuration.js
bantic: yeah. the default export object reference can't be changed, either, so if `configuration` is the default export then calls to `configure(newconfigObject)` do not change the configuration for consumers.

addon/providers/base.js
bantic: Not sure why this is here as it isn't used in the file. I can nix this line. jshint's "unused" is set to false because otherwise this PR would be chock full of jshint fixes.

index.js
bantic: agreed, this was a mistake, I'll nix it.

addon/configuration.js
mixonic: > Part of me trying to minimize diff noise was not changing existing var declarations.

:100:

zaccharles:
How's this going? :-)

mixonic:
@zaccharles PR will be update with any/all progress. We would love to free up the time to spend on this, happy to help a volunteer push it across the finish line though.

mitchlloyd:
Superseded by #281

PR #250, opened 2015-11-16
Iframe Service

jagthedrummer:
This is a freshly rebased branch of https://github.com/Vestorly/torii/pull/206

The basic overview, which is described in an update to the README is:

## Using an iframe instead of a popup

You can configure torii to use an in-page iframe instead of a separate
popup window for authentication. This can be done on either a global or
a per-provider basis.

To change this globally set the `remoteServiceName` variable in the main
torii config to be `'iframe'`.

```JavaScript
/* jshint node: true */
// config/environment.js
module.exports = function(environment) {
  var ENV = {
    /* ... */
    torii: {
      remoteServiceName: 'iframe',
      providers: { /* ... */ }
    }
  };
  return ENV;
};
```

If you only want the iframe for a single provider you can include the
`remoteServiceName` value in the configuration for that provider.

```JavaScript
/* jshint node: true */
// config/environment.js
module.exports = function(environment) {
  var ENV = {
    /* ... */
    torii: {
      // a 'session' property will be injected on routes and controllers
      providers: {
        'mycorp-oauth2': {
          remoteServiceName: 'iframe'
          /* ... */
        }
      }
    }
  };
  return ENV;
};
```

Once your provider has been configured you need to tell torii where to
apend the iframe when you call `session.open`.

For instance:

```JavaScript
this.get("session")
  .open("mycorp-oauth2",{ iframeParent : '#signin-modal-content'})
  .then(function(){
    // ...
  });
```

Of course, the element with the id of `signin-modal-content` needs to be
currently in the DOM.

/end of README excerpt

---

Instead of requiring the `iframeParent` argument to be passed to `open` we might include a component that people can include in a template to tell torii where to put the iframe. Something like:

```handlebars
{{torii-iframe-placeholder}}
```

I took a stab at making this happen, but so far I haven't been able to get the component to be useable in the host app. I suspect that this is related to the custom build system and the fact that I don't entirely understand how it works and how it's different from ember-cli based add-ons. Would love some pointers on that note.

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,6 +318,101 @@ via the `torii` property injected onto _routes_, or the `session` property
 injected onto routes and controllers (using the session management feature
 will require you to write an adapter for your application – see notes on session management below).
 
+## Using an iframe instead of a popup
+
+You can configure torii to use an in-page iframe instead of a separate
+popup window for authentication. This can be done on either a global or
+a per-provider basis.
+
+To change this globally set the `remoteServiceName` variable in the main
+torii config to be `'iframe'`.
+
+```JavaScript
+/* jshint node: true */
+// config/environment.js
+module.exports = function(environment) {
+  var ENV = {
+    /* ... */
+    torii: {
+      remoteServiceName: 'iframe',
+      providers: { /* ... */ }
+    }
+  };
+  return ENV;
+};
+```
+
+If you only want the iframe for a single provider you can include the
+`remoteServiceName` value in the configuration for that provider.
+
+```JavaScript
+/* jshint node: true */
+// config/environment.js
+module.exports = function(environment) {
+  var ENV = {
+    /* ... */
+    torii: {
+      // a 'session' property will be injected on routes and controllers
+      sessionServiceName: 'session',
+      providers: {
+        'mycorp-oauth2': {
+          remoteServiceName: 'iframe'
+          /* ... */
+        }
+      }
+    }
+  };
+  return ENV;
+};
+```
+
+Once your provider has been configured you need to tell torii where to
+append the iframe when you call `session.open` by using the
+`{{torii-iframe-placeholder}}` component. You need to make sure that
+this component is added to the DOM before you call `session.open` and if
+you give the user a way to back out of authentication (by closing a
+modal that contains the iframe, for instance) you need to make sure that
+the component is removed from the DOM so that torii will see that the
+auth flow has been cancelled.
+
+For instance, in `routes/application.js` you might have the following
+`signIn` action:
+
+```JavaScript
+signIn: function() {
+  var route = this;
+  // Set a value that will result in the placeholder component being
+  // added to the DOM
+  route.controller.set('signingIn',true);
+  // We need to user Ember.run.next to make sure that the placeholder
+  // component has been added to the DOM before session.open is called
+  Ember.run.schedule('afterRender', this, function(){
+    Ember.$('#signin-modal-back').one('click',function(){
+      route.controller.set('signingIn',false);
+    });
+    this.get("session")
+      .open("clickfunnels-oauth2")
+      .then(function(){
+        route.controller.set('signingIn',false);
+      });
+  });
+}
+```
+
+Then in `templates/application.hbs` you might have:
+
+```handlebars
+{{#if signingIn}}
+<div id="signin-modal-back">
+  <div id="signin-modal-frame">
+    <div id="signin-modal-content">
+      {{torii-iframe-placeholder}}
+    </div>
+  </div>
+</div>
+{{/if}}
+```
+
 ## Providers in Torii
 
 Torii is built with several providers for common cases. If you intend to
--- a/lib/torii/bootstrap/torii.js
+++ b/lib/torii/bootstrap/torii.js
@@ -35,13 +35,6 @@ export default function(application) {
   application.register('torii-service:iframe', IframeService);
   application.register('torii-service:popup', PopupService);
 
-  var remoteServiceName = configuration.remoteServiceName || 'popup';
-  if(remoteServiceName === 'iframe'){
-    application.inject('torii-provider', 'popup', 'torii-service:iframe');
-  }else{
-    application.inject('torii-provider', 'popup', 'torii-service:popup');
-  }
-
   if (window.DS) {
     var storeFactoryName = 'store:main';
     if (hasRegistration(application, 'service:store')) {
--- a/lib/torii/bootstrap/torii.js
+++ b/lib/torii/bootstrap/torii.js
@@ -9,9 +9,11 @@ import GithubOauth2Provider from 'torii/providers/github-oauth2';
 import AzureAdOauth2Provider from 'torii/providers/azure-ad-oauth2';
 import StripeConnectProvider from 'torii/providers/stripe-connect';
 import EdmodoConnectProvider from 'torii/providers/edmodo-connect';
+import configuration from 'torii/configuration';
 
 import ToriiService from 'torii/services/torii';
 import PopupService from 'torii/services/popup';
+import IframeService from 'torii/services/iframe';
 
 import { hasRegistration } from 'torii/lib/container-utils';
 
@@ -30,9 +32,15 @@ export default function(application) {
   application.register('torii-provider:edmodo-connect', EdmodoConnectProvider);
   application.register('torii-adapter:application', ApplicationAdapter);
 
+  application.register('torii-service:iframe', IframeService);
   application.register('torii-service:popup', PopupService);
 
-  application.inject('torii-provider', 'popup', 'torii-service:popup');
+  var remoteServiceName = configuration.remoteServiceName || 'popup';
+  if(remoteServiceName === 'iframe'){
+    application.inject('torii-provider', 'popup', 'torii-service:iframe');
+  }else{
+    application.inject('torii-provider', 'popup', 'torii-service:popup');
+  }
 
   if (window.DS) {
     var storeFactoryName = 'store:main';
--- a/lib/torii/mixins/ui-service-mixin.js
+++ b/lib/torii/mixins/ui-service-mixin.js
@@ -1,0 +1,128 @@
+import UUIDGenerator from 'torii/lib/uuid-generator';
+import PopupIdSerializer from 'torii/lib/popup-id-serializer';
+import ParseQueryString from 'torii/lib/parse-query-string';
+export var CURRENT_REQUEST_KEY = '__torii_request';
+
+var on = Ember.on;
+
+function parseMessage(url, keys){
+  var parser = ParseQueryString.create({url: url, keys: keys});
+  var data = parser.parse();
+  return data;
+}
+
+var ServicesMixin = Ember.Mixin.create({
+
+  init: function(){
+    this._super.apply(this, arguments);
+    this.remoteIdGenerator = this.remoteIdGenerator || UUIDGenerator;
+  },
+
+  // Open a remote window. Returns a promise that resolves or rejects
+  // accoring to if the iframe is redirected with arguments in the URL.
+  //
+  // For example, an OAuth2 request:
+  //
+  // iframe.open('http://some-oauth.com', ['code']).then(function(data){
+  //   // resolves with data.code, as from http://app.com?code=13124
+  // });
+  //
+  // Services that use this mixin should implement openRemote
+  //
+  open: function(url, keys, options){
+    var service   = this,
+        lastRemote = this.remote;
+
+    return new Ember.RSVP.Promise(function(resolve, reject){
+      if (lastRemote) {
+        service.close();
+      }
+
+      var remoteId = service.remoteIdGenerator.generate();
+
+      var pendingRequestKey = PopupIdSerializer.serialize(remoteId);
+      localStorage.setItem(CURRENT_REQUEST_KEY, pendingRequestKey);
+
+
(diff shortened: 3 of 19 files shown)


jagthedrummer:
@mixonic, I just realized that Friday managed to sneak past without me ever thinking about trying to sync up with you on this. Maybe sometime this week?

jagthedrummer:
Just pinging this again. :smiley:

lib/torii/bootstrap/torii.js
```diff
@@ -30,9 +32,15 @@ export default function(application) {
   application.register('torii-provider:edmodo-connect', EdmodoConnectProvider);
   application.register('torii-adapter:application', ApplicationAdapter);
 
+  application.register('torii-service:iframe', IframeService);
   application.register('torii-service:popup', PopupService);
 
-  application.inject('torii-provider', 'popup', 'torii-service:popup');
+  var remoteServiceName = configuration.remoteServiceName || 'popup';
+  if(remoteServiceName === 'iframe'){
```
mixonic: Can we do this lazily instead? Add a `popup` property on the base provider that looks up the proper thing based on the `config`?

mixonic: I think it would improve/remove the need for stuff in `walk-providers` as well. Instead of setting the popup, it should just inject a non-instantiated property with the name of the provider (or even with the config)

lib/torii/mixins/ui-service-mixin.js
```diff
@@ -0,0 +1,138 @@
+import UUIDGenerator from 'torii/lib/uuid-generator';
+import PopupIdSerializer from 'torii/lib/popup-id-serializer';
+import ParseQueryString from 'torii/lib/parse-query-string';
+//import {stringifyOptions, prepareOptions, parseMessage} from 'torii/helpers/services-helper';
```
mixonic: Can drop this if un-used

lib/torii/mixins/ui-service-mixin.js
```diff
@@ -0,0 +1,138 @@
...
+export var CURRENT_REQUEST_KEY = '__torii_request';
+
+var on = Ember.on;
+
+
```
mixonic: Some whitespace here :-)

lib/torii/mixins/ui-service-mixin.js
```diff
@@ -0,0 +1,138 @@
...
+
+function parseMessage(url, keys){
+  var parser = ParseQueryString.create({url: url, keys: keys}),
+      data = parser.parse();
```
mixonic: Please split these declarations into multiple statements, with commas they are tricky to drop a breakpoint on.

lib/torii/mixins/ui-service-mixin.js
```diff
@@ -0,0 +1,138 @@
...
+
+var ServicesMixin = Ember.Mixin.create({
+
+  init: function(options){
+    console.log('creating with options : ', options);
+    // may not need these next two lines : https://github.com/Vestorly/torii/commit/f5531d325b95df825ae763ef064c5acc36be8858#diff-8b7b6e15c618f576b36e529bef3c661bL51
+    this._super.apply(this, arguments);
+    options = options || {};
+    this.remoteIdGenerator = options.remoteIdGenerator || UUIDGenerator;
+    console.log('this.remoteIdGenerator = ', this.remoteIdGenerator);
```
mixonic: `console.log`

lib/torii/mixins/ui-service-mixin.js
```diff
@@ -0,0 +1,138 @@
...
+  return data;
+}
+
+
+
+var ServicesMixin = Ember.Mixin.create({
+
+  init: function(options){
+    console.log('creating with options : ', options);
+    // may not need these next two lines : https://github.com/Vestorly/torii/commit/f5531d325b95df825ae763ef064c5acc36be8858#diff-8b7b6e15c618f576b36e529bef3c661bL51
```
mixonic: `_super` would be needed, the `options` line can be dropped if you rewrite the logic for `options.remoteIdGenerator`

lib/torii/services/iframe.js
```diff
@@ -0,0 +1,29 @@
+import UiServiceMixin from 'torii/mixins/ui-service-mixin';
+
+var on = Ember.on;
+
+var Iframe = Ember.Object.extend(Ember.Evented, UiServiceMixin, {
+
+  // Open an iframe window.
+  openRemote: function(url, pendingRequestKey, options){
+    this.remote = Ember.$('<iframe src="'+url+'" id="torii-iframe"></iframe>'); //window.open(url, pendingRequestKey, optionsString);
```
mixonic: Comment unneeded

lib/torii/services/iframe.js
```diff
@@ -0,0 +1,29 @@
...
+    var iframeParent = 'body';
+    if(options && options.iframeParent){
```
mixonic: Space it out :-) `if (options && options.iframeParent) {`

Regardless, I think we should assert the option is passed instead of defaulting to `body`. It seems like a fairly useless default, just good for demos.

lib/torii/services/iframe.js
```diff
@@ -0,0 +1,29 @@
+import UiServiceMixin from 'torii/mixins/ui-service-mixin';
+
+var on = Ember.on;
+
+var Iframe = Ember.Object.extend(Ember.Evented, UiServiceMixin, {
+
+  // Open an iframe window.
```
mixonic: `Append an iframe`

test/tests/unit/services/popup-test.js
```diff
@@ -50,7 +50,7 @@ asyncTest("open resolves based on popup window", function(){
   var popupId = '09123-asdf';
   var mockWindow = null
 
-  popup = Popup.create({popupIdGenerator: buildPopupIdGenerator(popupId)});
+  popup = new Popup({remoteIdGenerator: buildPopupIdGenerator(popupId)});
```
mixonic: We should definitely use `create`. Invocation of Ember objects with `new` is not actually public API (I thought it was so if you see any left here, my fault :-p)

README.md
```diff
@@ -318,6 +318,70 @@ via the `torii` property injected onto _routes_, or the `session` property
...
+        }
+      }
+    }
+  };
+  return ENV;
+};
+```
+
+Once your provider has been configured you need to tell torii where to
+apend the iframe when you call `session.open`.
```
bantic: "append"

mixonic:
@jagthedrummer THANKS. This is looking really good man. Sorry it has taken to long to think it through and find the time.

I'd like to suggest one more largish adjustment. Right now, before an iframe is opened, the developer must ensure an element with the correct id (or selector) is on the page. This leaks some domish stuff where it doesn't feel appropriate.

Instead, let's add a component `{{torii-iframe}}` that can only be used once (should assert if used more than that) and can manage the DOM parts of the iframe.

lib/torii/bootstrap/torii.js
jagthedrummer: @mixonic If I'm following what you're saying here, would that mean that we want to do something like this here?

```javascript
var remoteServiceName = configuration.remoteServiceName || 'popup';
application.inject('torii-provider', 'remoteServiceProviderName',
  'torii-service:' + remoteServiceName);
```

And then we'd have a computed property in `lib/torii/providers/base.js` that somehow uses the value of the `'remoteServiceProvider'` property to look up either the popup or iframe service? If so, how do I look up a service that's registered on the application from inside a provider?

```javascript
  popup: computed('remoteServiceProviderName', function(){
    return ?.lookup(this.get('remoteServiceProviderName'));
  })
```

mixonic: Right, something along those lines. you can `inject` with `instantiate: false` if you want to just inject the remote service name directly as a prop.

jagthedrummer:
@mixonic @bantic In this branch/PR I've addressed all of the issues that you guys had pointed out with the exception of the "lazy registration" issue. I've worked on that and have something that is different, and I think a little more like what you're asking for. Unfortunately is doesn't actually work, and I'm not entirely sure why.

This commit is what I have going so far: https://github.com/jagthedrummer/torii/commit/64de9e1c532e5b88f03667789c9853c52efe140e

The issue is that even though the value of `'remoteServiceName'` is being set to `'iframe'` in `walk-providers`, when I actually call `get('popup')` on the provider, inside that computed property the value of `'remoteServiceName'` always comes back as `'popup'`.

One other issue I've discovered with this approach is that for the stuff in `walk-providers` to actually work for a custom provider you have to create an initializer for your provider that registers it with the application so that torii can find it. Registering the provider with the application isn't mentioned anywhere in the docs for creating a customer provider, so that would mean that any existing custom providers would have to be updated before they could be used with the iframe service. The "non-lazy" current code in this PR (not in the commit linked above) should be transparently backward compatible with any existing providers.

I feel like there's probably some way to do the "lazy" thing in a way that 1) actually works and 2) is transparently backward compatible with existing providers. What am I missing?

mixonic:
@jagthedrummer I'll attempt an implementation tomorrow and PR against your branch :-)

mixonic:
with lazy popup service discovery in https://github.com/jagthedrummer/torii/pull/1

jagthedrummer:
@mixonic Thanks for the PR on the lazy lookup. I just merged it into this branch/PR and everything is working perfectly in my test app. I think this should be all ready to go. Thanks for the help!

README.md
```diff
@@ -318,6 +318,101 @@ via the `torii` property injected onto _routes_, or the `session` property
...
+
+```JavaScript
+signIn: function() {
+  var route = this;
+  // Set a value that will result in the placeholder component being
+  // added to the DOM
+  route.controller.set('signingIn',true);
+  // We need to user Ember.run.next to make sure that the placeholder
+  // component has been added to the DOM before session.open is called
+  Ember.run.next(this,function(){
```
mixonic: `Ember.run.schedule('afterRender', this, function() {`, not `next`.

lib/torii/redirect-handler.js
```diff
@@ -21,7 +22,10 @@ var RedirectHandler = Ember.Object.extend({
         var url = windowObject.location.toString();
         windowObject.localStorage.setItem(pendingRequestKey, url);
 
-        windowObject.close();
+        var remoteServiceName = configuration.remoteServiceName || 'popup';
+        if(remoteServiceName === 'popup'){
```
mixonic: this seems bad. If the default config is `popup` but the specific running remote service is `iframe` the `close` would still be called.

lib/torii/services/iframe.js
```diff
@@ -0,0 +1,31 @@
+import UiServiceMixin from 'torii/mixins/ui-service-mixin';
+import UUIDGenerator from 'torii/lib/uuid-generator';
+
+var on = Ember.on;
+
+var Iframe = Ember.Object.extend(Ember.Evented, UiServiceMixin, {
```
mixonic: Should extend `Ember.Service`

lib/torii/services/iframe.js
```diff
@@ -0,0 +1,31 @@
...
+
+  // For some reason having this init in the UiServiceMixin will not work...
+  init: function(){
```
mixonic: should call `this._super(...arguments);`

lib/torii/services/iframe.js
```diff
@@ -0,0 +1,31 @@
+import UiServiceMixin from 'torii/mixins/ui-service-mixin';
+import UUIDGenerator from 'torii/lib/uuid-generator';
+
+var on = Ember.on;
+
+var Iframe = Ember.Object.extend(Ember.Evented, UiServiceMixin, {
+
+  // For some reason having this init in the UiServiceMixin will not work...
```
mixonic: This comment can maybe go away, or be improved? What is it supposed to tell a future traveler?

lib/torii/services/popup.js
```diff
@@ -40,120 +37,31 @@ function prepareOptions(options){
...
 }
 
-function parseMessage(url, keys){
-  var parser = ParseQueryString.create({url: url, keys: keys}),
-      data = parser.parse();
-  return data;
-}
-
-var Popup = Ember.Object.extend(Ember.Evented, {
+var Popup = Ember.Object.extend(Ember.Evented, UiServiceMixin, {
```
mixonic: Hm, can this also be a service? That they are objects is just old code from before services came in.

lib/torii/redirect-handler.js
jagthedrummer: Ah, yeah. Hmm... Seems like here in `redirect-handler` we need to somehow discover which provider we're handling and be able to find the config set for just that provider. Do you have a suggestion on how to approach that?

lib/torii/services/iframe.js
jagthedrummer: This was based on the existing code for `services/popup.js` which does not extend `Ember.Service`. Should both of them be updated to extend that?

lib/torii/services/iframe.js
jagthedrummer: Does `services/popup.js` need this also? https://github.com/Vestorly/torii/blob/master/lib/torii/services/popup.js#L51-L53

lib/torii/services/iframe.js
jagthedrummer: Yeah, it could be better. It's basically supposed to mean "even though this identical code shows up in both popup.js and iframe.js DON'T try to move it to the UiServiceMixin that they both inherit from".

PR #256, opened 2015-12-29
Sending the `transition` as a parameter with the `accessDenied` action

bl4ckm0r3:
Having the `transition` sent through the `accessDenied` action makes it much easier to redirect to the desired route after completing the normal signIn flow.


mixonic:
:100: Thank you!

bl4ckm0r3:
you are welcome :+1:  thank you!

mwpastore:
I needed this today and lo' and behold here it is! Thank you!

Issue #257, opened 2016-01-04
Timing issues with iframe + liquid fire modal


jagthedrummer:
This may be more of a liquid fire issue than a torri issue, but it might affect what we want to say in the README.

I'm using a liquid fire modal to host the `torii-iframe-placeholder` component, and using `Ember.run.schedule('afterRender'...` to call `session.open` often results in a blank modal in Safari, both on desktop (OS X) and on various iOs devices. The issue seems to be that at the time `session.open` is called there are actually two instances of the `torii-iframe-placeholder` component in the DOM. And it seems that the one the iframe is appended to is removed shortly thereafter, leaving the other placeholder in the modal, but with no iframe.

Here's my `signIn` action:

```javascript
    signIn: function() {
      var _this = this;

      console.log( $('.torii-iframe-placeholder').length ); // length is 0

      // This next line triggers Liquid Fire to show the sign-in modal
      this.controller.set('signIn','');
      Ember.run.schedule('afterRender',this,function(){
        console.log( $('.torii-iframe-placeholder').length ); // length is 2
        this.get("session")
          .open("clickfunnels-oauth2-bearer")
          .then(function(data){
            _this.controller.set('signIn',null); // remove the modal
            _this.send('afterSignIn',data);
          },function(error){
            console.log('signIn failed', error);
          });
      });
    }
```

After the blank modal is shown, if I run `$('.torii-iframe-placeholder').length` in a console I can see that there is only one instance.

As a work around I've found that if I use `Ember.run.later` with a timeout of 500 milliseconds then things seem to work consistently. I'd guess that it works due to lucky timing, and that there could still be instances (device under heavy load) where even 500 ms is not long enough for the DOM to have settled down.

Any ideas on the right way to solve this?

jagthedrummer:
Now I'm realizing that maybe I need to have my modal component emit an event after it's been rendered, and then start the torii process as a result of that event. I'll try that and update this issue.

jagthedrummer:
Welp, that approach isn't working either. When I emit an event from my modal component on either `didInsertElement` or `didRender` then in the console I get 3 copies of the deprecation warning below.

Additionally, in Safari it looks like the `torii-iframe-placeholder` component (which is embedded in the template for my modal component) has not yet been added to the DOM by the time the `didInsertElement` or `didRender` hooks are fired on the parent component. Is there something like `childrenDidRender`? Or maybe this is entirely the wrong approach?

---------
[...]

PR #264, opened 2016-01-18
Send an action when the torii-iframe-placeholder component has rendered.

jagthedrummer:
This allows the app to avoid any timing issues that may result from the
placholder component being shown (as reported in issue #257).


jagthedrummer:
Is there anything I can do to help this along?

jagthedrummer:
Ping @bantic, @mixonic. Are there concerns with this PR and/or approach?

bantic:
@jagthedrummer Sorry for the delay. This makes sense to me. Is it possible to add an integration test that the event is emitted?

Issue #262, opened 2016-01-15
Oauth2-code doesn't pass response params in Authorization Object


averydev:
The OAuth provider I'm trying to use (Shopify) requires usage of response params other than `authorizationCode`, `provider`, and `redirectUri` to validate the response prior to requesting a permanent token.  

It seems like this should add the responseParams requested by the user to the `authorization` object for later use.

Is there a reason that these properties aren't included? If not, I'm happy to submit a PR...

mattmarcum:
:+1: I just spent an hour digging through the code to see why I wasn't getting any of my response params passed to my custom provider.  Looking into https://github.com/Vestorly/torii/blob/master/lib/torii/providers/oauth2-code.js#L149  I see that `oauth2-code:open` does check for missing `responseParams` but it never does nothing with them after that.  It should probably be sticking them on the object that is returned on line 166.  

You can switch to extending `oauth2-bearer` and that will get you all the responseParams returned under `authorizationToken` as an object with key/value pairs based on responseParams.  https://github.com/Vestorly/torii/blob/master/lib/torii/providers/oauth2-bearer.js#L37

That works, but I'm not sure if that is what is really supposed to happen here.  Based on the key name of `authorizationToken` you'd expect it to be a string, not an object.  I think a better solution would be just to stick a new entry on the returned object with key of `responseParams` and with the value set to the `responseParms` object.

I think docs for implementing a custom provider that extends the oauth2-* base classes would be nice.

PR #277, opened 2016-02-11
Add Response params to the authorize object for Oauth2

averydev:
Including `responseParams` to an oauth2 provider will return an error if those params are missing from a response (as it should). However those responses are not passed back along. Those responses are generally to either verify that the server response is valid, or for further use in requesting a token. 

This adds those requested params to the authorization object.

### DIFF
--- a/lib/torii/providers/oauth2-code.js
+++ b/lib/torii/providers/oauth2-code.js
@@ -134,42 +134,44 @@ var Oauth2 = Provider.extend({
    * If there was an error or the user either canceled the authorization or
    * closed the popup window, the promise rejects.
    */
-  open: function(options){
-    var name        = this.get('name'),
-        url         = this.buildUrl(),
-        redirectUri = this.get('redirectUri'),
-        responseParams = this.get('responseParams'),
-        responseType = this.get('responseType'),
-        state = this.get('state'),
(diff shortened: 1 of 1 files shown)


lib/torii/providers/oauth2-code.js
```diff
@@ -134,42 +134,44 @@ var Oauth2 = Provider.extend({
...
+        authObj = {
+          authorizationCode: authData[responseType],
+          provider: name,
+          redirectUri: redirectUri
+        };
+
+        responseParams.forEach(function(param){
+          if (authData[param] === undefined) {
+            missingResponseParams.push(param);
+          }else if(param !== responseType){
```
mixonic: Perhaps this check should not be in place? It would mean the `authData[responseType]` is passed back twice, once as the normalized `authorizationCode` and once as some other un-normalized value.

That seems ok to me. Thoughts?

whatthewhat:
@mixonic 👋 
We've just hit this issue while implementing a custom provider. Would you be open to merging an updated version of this PR?

I'm thinking it might be a little cleaner to include the entire `authData` object as the 4th key (`responseParams`?) as mentioned in https://github.com/Vestorly/torii/issues/262#issuecomment-179586734

PR #280, opened 2016-02-25
Then-able hook in facebook-connect. Fixes #279

### LINKED ISSUE #279: FB is undefined on session fetch
I'm using the facebook-connect provider and I've noticed that FB is sometimes not defined on app boot (full page refresh). I think I understand the problem but I can't quite get a handle on a solution. Here's the normal login flow:

1. checkLogin calls session fetch
1. session fetch fails because user is not logged in (no valid JWT)
1. user is redirected to login screen
1. user clicks login
1. session open called with facebook-connect provider
1. fbLoad() fires and defines FB
1. user logs in with Facebook (or is already logged in)
1. application adapter makes some housekeeping calls* to FB.api()

Here's the refresh flow:

1. checkLogin calls session fetch
1. session fetch succeeds because user is logged in (valid JWT)
1. session fetch also validates the result of fb.getLoginStatus() to make sure Facebook access is still good
1. application adapter makes some housekeeping calls* to FB.api()

The facebook-connect adapter is never invoked in the refresh so fbLoad() never fires and defines FB. I can't understand why the refresh flow works only some of the time; you'd think FB would never be defined and it would always fail. Am I right about what's happening here or is something else going on?

Short of injecting the torii service into my application adapter and invoking torii open with the facebook-connect provider if FB isn't defined, is there another way to solve this problem? Thanks in advance.

* - specifically, retrieving the user's current Facebook permissions

mwpastore:


lib/torii/providers/facebook-connect.js
```diff
@@ -97,13 +98,17 @@ var Facebook = Provider.extend({
...
+  loadFb: function(){
+    return fbLoad( this.settings() );
+  },
+
   // Load Facebook's script eagerly, so that the window.open
   // in FB.login will be part of the same JS frame as the
   // click itself.
   loadFbLogin: on('init', function(){
-    fbLoad( this.settings() );
+    this.loadFb();
```
mwpastore: The "problem" is that `loadFbLogin()` doesn't return fbPromise, so you can't chain off of it. This "solution" just adds an intermediate method, `loadFb()`, that _does_ return fbPromise. You're correct, the `loadFbLogin()` just does what it did before because I didn't want to break anything; so yes, my goal was to add a new public API without duplicating code or causing unintended consequences. 

I'm using it like this:

```javascript
  fetch() {
    let fbProvider = Ember.getOwner(this).lookup('torii-provider:facebook-connect');

    return fbProvider.loadFb().then(() => {
      // now I can call FB.getLoginStatus() or what have you
```

Yes, if there were a way to delay the execution of my application adapter's `fetch()` hook until the provider (and the FB SDK) is fully loaded, that would be ideal! I just had no idea how to accomplish that or if it was even an option.

mixonic: We have the code, we can make it happen :muscle: ;-)

I would strongly prefer the introduction of an optionally defined `load` method on providers that would be called before open, fetch, and close by the session.

Actually you kind of want it to be run before the provider's `open` as well, but that is unfortunately not permitted in our current API design :-/ since `open` is defined by the implementation and called directly by the consumer.

@mwpastore if you can push it forward I'm happy to chat and keep it making progress with you. But I definitely believe a general solution is better here than an ad-hoc API addition.

PR #281, opened 2016-02-26
convert to ember-cli

mitchlloyd:
Based on #244 and freshly rebased.

Squashed commits:

  * Add ember-try for emer 1.13.7, 2.0, 2.1, beta, canary
  * add ember 1.12.1 to ember-try
  * remove `isDevelopingAddon`, ember-addon/index, unused var get
  * Post rebase fixes
  * Remove old files

To do's carried over:
  * [ ] ~~update this or make a separate PR to remove the old grunt build system~~ Separate PR -- Let's prioritize merging to avoid more rebasing.
  * [x] real-world testing -- Did a quick test against real application using Torii. This application uses a custom adapter and provider to perform simple authentication with email / password, returning a session key. @bantic Might be good if you could try something out as well.
  * [x] use ember-try to test different ember versions
  * [ ] ~~update documentation to point out that globals-mode Torii will no longer be supported when this is an ember-cli addon~~ Let's do this in a separate PR if possible.


bantic:
Tested on a bunch of local projects, have not seen any issues. This is great! Finally, ember-cli! :)

PR #288, opened 2016-04-12
Cleanup pending requests

davewasmer:
Currently, Torii tracks in flight requests using a localStorage key. If the browser window is closed, that key remains behind. This results in Torii thinking there is an in flight request the next time it runs on that same domain. It then attempts to close the opened remote, which fails:

image

The solution implemented here uses the browser's `onbeforeunload` hook to remove the pending request key from localStorage, ensuring a clean state for the next page load.

## Steps to reproduce:

1. Open a torii session that results in a popup or iframe opening.
2. Close the parent browser window without touching the remote (iframe/popup) window.
3. Navigate back to the page in a new window, and attempt to open another torii session.

### DIFF
--- a/addon/mixins/ui-service-mixin.js
+++ b/addon/mixins/ui-service-mixin.js
@@ -47,6 +47,14 @@ var ServicesMixin = Ember.Mixin.create({
       service.openRemote(url, pendingRequestKey, options);
       service.schedulePolling();
 
+      var onbeforeunload = window.onbeforeunload;
+      window.onbeforeunload = function() {
+        if (typeof onbeforeunload === 'function') {
+          onbeforeunload();
+        }
+        service.close();
+      };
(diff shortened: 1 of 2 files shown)


mixonic:
@davewasmer this seems good. I'm brainstorming if there is another way to do this though- it seems bad that we would trust a token from a previous HTML document render.

davewasmer:
@mixonic perhaps wiping that particular local storage key in the initializer?

mixonic:
A solution would also need to account for multiple tabs being open at the same time. I know this current solution doesn't, but multi-tab is a constraint to consider regardless.

davewasmer:
Ah - sounds like maybe the local storage key needs to be unique per window context (i.e. unique across tabs and reloads)?

PR #292, opened 2016-04-25
update broccoli deps

stefanpenner:
These are just the ones consumers of this add-on would get. Im sure some of the devDeps could also be updated but that is likely more involved.


mixonic:
😱 seems good!

Issue #296, opened 2016-05-18
Error when FB is already assigned to the window object


andorfermichael:
Hi,

I am currently developing an ember application which uses the facebook-connect provider for authentification and ember-simple-auth for session handling.
I have spent hours on searching for an error because when I pressed my login button simply nothing had happened (neither an error message).
However, finally I found out that FB was initialized and assigned to the window object twice which stopped Torii from authenticating.
By applying a hack in node_modules/torii/lib/torii/providers/facebook-connect.js I could solve the problem:

```
function fbLoad(settings){
  if (fbPromise) { return fbPromise; }

  var original = window.fbAsyncInit;
  var locale = settings.locale;
  delete settings.locale;
  fbPromise = new Ember.RSVP.Promise(function(resolve, reject){
    if (window.FB) {
      return resolve()
    }
    window.fbAsyncInit = function(){
      FB.init(settings);
      Ember.run(null, resolve);
    };
    $.getScript('//connect.facebook.net/' + locale + '/sdk.js');
  }).then(function(){
    window.fbAsyncInit = original;
    if (window.fbAsyncInit) {
      window.fbAsyncInit.hasRun = true;
      window.fbAsyncInit();
    }
  });

  return fbPromise;
}
```

I simply prove if FB is already assigned to the window object and if so I resolve the promise.

I think this should be fixed within Torii since such a hack is neither available to other developers who clone the project nor to me if I will run npm install once again and I think it will affect other developers too.

PR #297, opened 2016-05-19
Implement condition which checks if FB was already initialized and assigned to the window object

andorfermichael:
If FB was already initialized and assigned to the window object by another service, torii stops authentication but will not throw any error or warning. 

The added condition prevents from this error.

For further information please see [issue #296](https://github.com/Vestorly/torii/issues/296).

### DIFF
--- a/addon/providers/facebook-connect.js
+++ b/addon/providers/facebook-connect.js
@@ -18,6 +18,7 @@ function fbLoad(settings){
   var locale = settings.locale;
   delete settings.locale;
   fbPromise = new Ember.RSVP.Promise(function(resolve, reject){
+    if (window.FB) { return resolve(); }
     window.fbAsyncInit = function(){
       FB.init(settings);
       Ember.run(null, resolve);


mixonic:
@andorfermichael it seems bad that the `settings` you pass in this case are completely ignored. Should this perhaps be an assertion instead of a fix?

andorfermichael:
@mixonic Ah, you are right, I did not consider that settings were set there. Maybe an assertion would be better but wouldn't there be a way to inject the settings in the existing FB instance? Since the goal is to get running torii and other tools which create and/or use FB . And how would an assertion help in this case?

mixonic:
The assertion would ensure this troublesome case is caught and brought to the user's attention more clearly. It definitely is not a fix.

I think a good fix would be a way to pass `window.FB` into the `open` command as a param. "bring your own `FB`", and if you passed it, use that instead of generating one. Does that seem plausible?

andorfermichael:
@mixonic yes, this seems absolutely plaubsible. If I use torii in combination with ember-simple-auth then we also have to adapt ember-simple-auth haven't we? Because we would have to pass the window.FB in here if I got that right: 

return this
        .get('session')
        .authenticate('authenticator:torii', 'facebook-connect', window.FB)
        .then(() => {
          this.fetchCurrentUser()
            .then(() => this.transitionTo('world'))
            .catch(err => console.error(err));
        })

mixonic:
@andorfermichael hm, I'm not really familiar with the ESA integration. I expect yes, you would need a way to pass an option through to Torii. However I expect that already exists?

I would definitely suggest passing an `options` object instead of positional arguments.

```js
let toriiProviderOptions = { FB: window.FB };
return this.get('session'.authenticate('authenticator:torii', 'facebook-connect', toriiProviderOptions).then(() => {
  return this.fetchCurrentUser() /* ... */
```

bantic:
This was caught on a conflict so I rebased it in #375

PR #298, opened 2016-05-20
Add Microsoft Live Oauth2 provider

paulnice:


YoranBrondsema:
@paulnice I think it's better to add this provider as a separate addon, for 2 reasons:

* It avoids adding code to `torii` that is not used by all users, thereby keeping it lightweight.
* You're not dependent on the maintainers of this repo to get your code published :-).

You can find examples of such addons on the [third-party addons section](https://github.com/Vestorly/torii#third-party-addons) of the readme.

Issue #307, opened 2016-08-18
May I use my sessionServiceName with the session service stub?


yethon:
I like how torii lets me define my own session service name 😄 

``` es6
 torii: {
      // a 'session' property will be injected on routes and controllers
      sessionServiceName: 'yethons-sweet-session',
      providers: {
        'facebook-oauth2': {
          apiKey:      'facebook-app-id',
          redirectUri: '/my-custom-landing-uri' // default is the current URL
        }
      }
    }
```

However, the session service stub assumes that I called my session `service` 😕 
`torii/test-support/helpers/torii.js`

``` es6
var session = application.__container__.lookup('service:session');
```

And that means I have to create my own helper...

bantic:
@yethon Thanks for raising, and for the PR. :)

PR #308, opened 2016-08-18
Genericize torii session name.

yethon:
This PR would make the torii session helper get the `sessionServiceName` that the user has defined.

Fixes the issue: [https://github.com/Vestorly/torii/issues/307](https://github.com/Vestorly/torii/issues/307)


bantic:
@yethon Thanks for this!

PR #311, opened 2016-08-29
Add ember-cli blueprints for torii adapters.

w-hat:


bantic:
@w-hat Thanks for this! Blueprints for Torii would be handy. Could you change boilerplate for the adapter to include:

  * `open`, `fetch` and `close` hooks
  * link in a comment to the Adapter docs: https://github.com/Vestorly/torii#adapters-in-torii

As far as a blueprint for a torii provider goes, that is the one that only needs the `open` hook implemented. By default it should use `Ember.Object.extend({ open() { ... } });`, but if you wanted to add a `--base=oauth2-code` option that inherits from [`torii/providers/oauth2-code`](https://github.com/Vestorly/torii/blob/master/addon/providers/oauth2-code.js) that would be cool.

thanks!

Issue #320, opened 2016-11-17
Facebook login (Oauth2) fails form in-app Safari


jdurand:
Everything works fine when logging in from the Safari browser on iOS, but doing the same from Facebook’s in-app browser fails with "remote could not open or was closed".

I debugged the situation and it turns out that `window.open()` doesn’t return a [WindowObjectReference](http://www.gtalbot.org/BugzillaSection/Bug195867GDR_WindowOpen.html) when executed from in-app Safari (Facebook): https://github.com/Vestorly/torii/blob/master/addon/services/popup.js#L45

As far as I can tell, the login process needs that reference to know if and when to close the window.
Has anyone found a way around this?

Also, note that the iframe solution is not an option as facebook doesn't allow their login form to be displayed inside an iframe.

jdurand:
The only way I found to make this work is to support a full page redirection from Facebook:

````
handleLoginRedirect() {
  let code = getUrlParameter('code');
  let state = getUrlParameter('state');
  let provider = 'facebook';

  Ember.run.schedule('afterRender', () => {
    let adapter = Ember.getOwner(this).lookup(`torii-adapter:${provider}-oauth2`);
    let state = this.get('authSession.stateMachine');

    state.send('startOpen');
    adapter.open({
      authorizationCode: code,
      redirectUri: `${window.location.protocol}//${window.location.host}${window.location.pathname}`
    }).then((payload) => {
      state.send('finishOpen', payload.user);
    }).catch((error) => {
      state.send('failOpen', error);
    });
  });

  return true;
}
````

You’ll also need to configure the proper **Valid OAuth redirect URIs** setting in the Facebook Login settings of your app.

I scratched my head for a few hours on this so I figured this could help others not spend as much time as I did on this.

jeroenhouben:
@jdurand where do you put this code?

I think I have the same problem, but from an iOS webapp (add2home). Things work fine in Safari but in add2home (I call this "webapp". it's when the navbar etc are hidden and the URL has it's own homescreen icon) it shows a blank page.

I can inspect the page and can see in a JS console that the URL is something with the authCode appended. So login worked, but torii is not picking up on the changes or something..

jdurand:
@jeroenhouben I also experienced this issue in _fullscreen webapp_ mode; it looks like it's the same issue.

Where you need to put this code depends on your application logic; I have a session service that calls this method on `init()`.

[...]

PR #321, opened 2016-11-18
Fixes issue with Facebook in-app browser not handing window.open

jdurand:
Facebook’s in-app browser doesn't play nice with `window.open()`... it navigates to the URL in the current tab and returns `undefined` instead of passing a window reference.
This doesn't completely fix the issue, it simply avoids setting the state machine in failed state when in fact it is loading the login page in the current "tab".

This fix is related to https://github.com/Vestorly/torii/issues/320


wpjmurray:
@jdurand Did you end up finding a better solution for this issue or was it resolved with a newer version of torii? I'm experiencing the same issue when trying to authenticate via Facebook or Google from the Facebook in-app webview. Any help you can provide would be appreciated. Thanks.

jdurand:
I switched to using Facebook Connect provider and I stopped having this issue. I'm guessing there is (or was) an issue with the underlying OAuth2 provider.

PR #324, opened 2016-12-01
App torii session should require addon torii session

topaxi:
`app/services/torii-session` tried to require `torii/services/session` which does not exist.
This triggered a warning in my console output.
I assume it should be requiring `torii/services/torii-session`


bantic:
@topaxi Thanks. I'm curious — how did you trigger the warning? Did you do `session: Ember.inject.service('torii-session')` somewhere?

bantic:
released in v0.9.3

PR #326, opened 2016-12-07
fix query string in last versions Ember.js

enniel:


mixonic:
@enniel this makes sense, good catch 👍 

Hm unsure what these failures on 2.11 and 2.12 are though

mixonic:
Hm I think those issues are related to 2.10 as well.

mixonic:
I updated master to test 2.10 release. The error currently raised was added in https://github.com/emberjs/ember.js/commit/22e1612ad07249fddb6a1e0bb2344eba638fab46#diff-162a95b4162d14de7e38fd4debc5fbd4R46 and perhaps is only related to the testing harness.

enniel:
I'm fixed test file.

enniel:
Any progress on this?

bantic:
@enniel thank you for this work! The failing tests were addressed by two other PRs that were just merged (via #344). I attempted a cherry-pick of the InertRenderer in #345 but those tests didn't pass against Ember 2.8.

PR #336, opened 2017-03-09
Update year in main callout text

elbeezy:
2016 -> 2017


balinterdi:
Build broken because of https://github.com/Vestorly/torii/pull/340 . Merging in that PR would probably fix this one, too.

bantic:
@elbeezy Thanks for the update!

PR #322, opened 2016-11-21
Replace `Ember.K` usage with JavaScript syntax

locks:

PR #338, opened 2017-03-16
fix Ember.K deprecation warning

kturney:
http://emberjs.com/deprecations/v2.x/#toc_code-ember-k-code


kturney:
looks like this was already and better addressed by https://github.com/Vestorly/torii/pull/322

PR #341, opened 2017-03-24
Handle frozen url params properties

### LINKED ISSUE #337: QueryString.requiredParams is frozen
We are running into this issue in the Ember Simple Auth dummy app: simplabs/ember-simple-auth#1260 currently and the reason seems to be that the `QueryString`'s `requiredParams` property is frozen. It also seems to relate to this issue in Ember.js itself emberjs/ember.js#15020 but the fact that the property is frozen in the first place seems a bit strange as well. It doesn't seem like torii actually freezes anything and this might very well not be a problem in torii so feel free to close this and see this more as a notification that there's something strange going on.

marcoow:
Ember will freeze concatenated properties so they cannot be mutated once applied. While that makes sense, torii's `QueryString` does not currently work when passed frozen `requiredParams` or `optionalParams` as frozen arrays cannot be passed to `Ember.A` (see discussion here: emberjs/ember.js#15020). This fixes that by simply cloning those properties before passing them to `Ember.A`.

Closes #337

### DIFF
--- a/addon/lib/query-string.js
+++ b/addon/lib/query-string.js
@@ -35,8 +35,8 @@ function getOptionalParamValue(obj, paramName){
 export default Ember.Object.extend({
   init: function() {
     this.obj               = this.provider;
-    this.urlParams         = Ember.A(this.requiredParams).uniq();
-    this.optionalUrlParams = Ember.A(this.optionalParams || []).uniq();
+    this.urlParams         = Ember.A(this.requiredParams.slice()).uniq();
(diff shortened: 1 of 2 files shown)


bantic:
This was handled by #344. Thanks for the fix, @marcoow !

bantic:
released (via #344) in v0.8.2 — @marcoow Is your issue fixed when using Torii @ 0.8.2 ?

marcoow:
Yep, works with 0.8.2!

PR #342, opened 2017-03-24
fix Circle CI build

marcoow:
The ember-try config is actually broken currently which breaks the CircleCLI build…

This also updates the node version used on Circle to 4 from 0.10.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,6 @@
 machine:
   node:
-    version: 0.10.28
+    version: "v4"
 dependencies:
   pre:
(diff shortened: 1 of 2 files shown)


marcoow:
The build is running now but failing. It seems that these failures should be fixed by #341 though. Not sure how you want to handle that (merge this first although the build is failing and then rebase #341?)

bantic:
@marcoow thanks! I'll merge this now and look at rebasing #341 as you suggest. Thanks for taking the time to update this.

PR #343, opened 2017-03-26
Replace Ember.K with a function that returns this

drewchandler:
`Ember.K` is deprecated (http://emberjs.com/deprecations/v2.x/#toc_code-ember-k-code).

I've replaced all instances of `Ember.K` with a function that returns `this`. It seems like an empty function would do the trick too, but returning `this` gives the exact same behavior as before.


drewchandler:
Sorry! Didn't see that https://github.com/Vestorly/torii/pull/322 already existed.

PR #345, opened 2017-03-27
cherry-pick fix test: use InertRenderer in test

bantic:
Uses the InertRenderer in unit test that instantiates a component

Cherry-picks one of the commits from #326


bantic:
Hmm, looks like the InertRenderer fails in 2.8. I will close this for now and keep the DummyRenderer as is.

Issue #294, opened 2016-05-11
remove grunt


bantic:
Torii is now fully an ember-cli addon and it no longer needs to use `grunt` to build.

PR #347, opened 2017-03-28
Remove Grunt

bantic:
After converting Torii to an ember addon (#281), grunt isn't really used anymore. It was used to create a globals build of Torii but that hasn't worked since v0.8.0 and no one has complained, and it was also used to generate yuidocs but we can bring that back without grunt.


bantic:
fixes #294

PR #354, opened 2017-05-08
Ember 2.13 - Fix deprecated usage of router.router

Dhaulagiri:
https://emberjs.com/deprecations/v2.x/#toc_ember-router-router-renamed-to-ember-router-_routermicrolib

PR #357, opened 2017-05-12
Fix this.router deprecation

t4t5:
Using `this.router` is deprecated and gives warnings starting from Ember 2.13 (See https://emberjs.com/deprecations/v2.x/#toc_ember-router-router-renamed-to-ember-router-_routermicrolib).

This change migrates to the new `_routerMicrolib` while staying backwards compatible.

### DIFF
--- a/addon/router-dsl-ext.js
+++ b/addon/router-dsl-ext.js
@@ -12,6 +12,8 @@ Router.reopen({
   _initRouterJs: function() {
     currentMap = [];
     this._super.apply(this, arguments);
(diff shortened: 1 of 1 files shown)


ryanlitalien:
@t4t5 CircleCI failure

Dhaulagiri:
@t4t5 this is a duplicate of #354 and, according to something @mwpastore pointed out, doesn't address all the deprecation warnings.  I'm going to close this in favor of pulling everything together into #354.

PR #367, opened 2017-05-24
Move examples into test dummy app

bantic:
Remove the `examples/` top-level dir that is not currently used, pull those same examples into the test dummy app.

Also remove some unused broccoli-related dependencies.


bantic:
released in v0.9.3

PR #370, opened 2017-05-25
Fix redirectUri regression and add disableRedirectInitializer option

sukima:
* Fixes a regression introduced by commit f0b214f ("Deprecate Redirect Handler, use public/redirect.html as redirect target", 2017-05-22).

* Adds a way to disable the deprecated behavior. This caused a bug where attempting to refresh a Torii enabled app would cause the second instance to freeze because it notices the localStorage key and attempts to close the main window and then crashes. A second refresh will work because the localStorage key was removed before the crash. And a third refresh would be back in the same boat. The patterns continues on. Since we have a new `redirect.html` instead this behavior can be switched off via a configuration flag until the deprecated code is removed in a future release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,6 +406,10 @@ By default Torii sets the `redirectUri` to
 `<currentURL>/torii/redirect.html`. If you wish to use the deprecated behavior
 then you will also have to manually configure the `redirectUri` to be `/`.
 
+If you are no longer relying on the deprecated behaviour and wish for it to no
+longer be executed you can manually disable it by setting
+`disableRedirectInitializer` to false in your `config.environment.js`.
+
 ## Providers in Torii
 
 Torii is built with several providers for common cases. If you intend to
--- a/addon/redirect-handler.js
+++ b/addon/redirect-handler.js
@@ -9,6 +9,13 @@
 import { CURRENT_REQUEST_KEY, WARNING_KEY } from "./mixins/ui-service-mixin";
 import configuration from 'torii/configuration';
 
+export class ToriiRedirectError extends Ember.Error {
+  constructor() {
+    super(...arguments);
+    this.name = 'ToriiRedirectError';
+  }
+}
+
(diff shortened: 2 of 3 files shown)


sukima:
@bantic / @elidupuis Here is the fix for the regression and a way to disable the initializer. Maybe there is a better way?

bantic:
@sukima Thanks for putting this together. I'd prefer to turn the initializer into a no-op rather than add this early-reject into the redirect handler. I'm going to cherry-pick your first commit into its own PR and merge it.
Do you want to modify the disable redirect option so that it turns the initializer function here into an early return instead? https://github.com/Vestorly/torii/blob/master/app/initializers/initialize-torii-callback.js#L7

sukima:
I can. I only hesitated doing so because the initializer didn't have the config imported.

sukima:
@bantic updated!

README.md
```diff
@@ -406,6 +406,10 @@ By default Torii sets the `redirectUri` to
 `<currentURL>/torii/redirect.html`. If you wish to use the deprecated behavior
 then you will also have to manually configure the `redirectUri` to be `/`.
 
+If you are no longer relying on the deprecated behaviour and wish for it to no
+longer be executed you can manually disable it by setting
+`disableRedirectInitializer` to false in your `config.environment.js`.
```
elidupuis: Should this say `true` instead of `false`?

bantic: @elidupuis Yeah, that should say `false`.
@sukima Also, you want to import `getConfiguration` (like the [torii session initializer does](https://github.com/Vestorly/torii/blob/606334ca1832ecadc75628f8b6a2c4e110c1e642/app/initializers/initialize-torii-session.js)) rather than `configuration` directly. I still have some time this morning to work on Torii so I can fix this if you don't have the time.

bantic:
closing for #374

PR #371, opened 2017-05-25
Make redirectUri configurable again

bantic:
This is a regression introduced by commit f0b214f ("Deprecate Redirect
Handler, use public/redirect.html as redirect target", 2017-05-22).

Supersedes part of #370

cc @sukima


bantic:
published in v0.9.4

PR #333, opened 2017-02-14
Google OAuth2 provider update

jfschaff:
Hi, this fixes what I mentioned in issue #331.

I implemented a new provider, named 'google-oauth2-bearer-v2', which uses Google's most recent authentication URL. It also performs an extra step: once a token is received, it is first validated by a second request to another URL (see [that link](https://developers.google.com/identity/protocols/OAuth2UserAgent#tokeninfo-validation) for more details).

I tested this provider with Ember Simple Auth, and fixed a couple of initial issues. I also included an example use of this provider in the example application, and instructions on how to set the provider.

PR #372, opened 2017-05-25
Added new provider 'google-oauth2-bearer-v2'.

bantic:
New provider for Google. This provider takes care of validating the token before declaring the authentication complete.

Based on #333, thanks @jfschaff!


bantic:
released in v0.9.5

PR #373, opened 2017-05-25
removed current request key from local storage if popup fails to open

### LINKED ISSUE #249: Prevented popup window's CURRENT_REQUEST_KEY should be removed from localStorage
In the Torii popup service, if the popup fails to load the `CURRENT_REQUEST_KEY` is not removed for `localStorage`, causing subsequent loads to attempt to close a non-existing window, which in turn causes the browser to issue a warning similar to `Scripts may close only the windows that were opened by it.`.

https://github.com/Vestorly/torii/blob/master/lib/torii/services/popup.js#L84
https://github.com/Vestorly/torii/blob/master/lib/torii/redirect-handler.js#L24

bantic:
fixes #249

### DIFF
--- a/addon/mixins/ui-service-mixin.js
+++ b/addon/mixins/ui-service-mixin.js
@@ -61,6 +61,7 @@ var ServicesMixin = Ember.Mixin.create({
       if (service.remote && !service.remote.closed) {
         service.remote.focus();
       } else {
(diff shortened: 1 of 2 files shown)


bantic:
released in v0.9.5

PR #374, opened 2017-05-25
add disableRedirectInitializer option

bantic:
from @sukima's work on #370:

* Adds a way to disable the deprecated behavior. This caused a bug where attempting to refresh a Torii enabled app would cause the second instance to freeze because it notices the localStorage key and attempts to close the main window and then crashes. A second refresh will work because the localStorage key was removed before the crash. And a third refresh would be back in the same boat. The patterns continues on. Since we have a new redirect.html instead this behavior can be switched off via a configuration flag until the deprecated code is removed in a future release.

Adding `disableRedirectInitializer` to the torii config skips initializing the redirect handler altogether.

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,6 +406,10 @@ By default Torii sets the `redirectUri` to
 `<currentURL>/torii/redirect.html`. If you wish to use the deprecated behavior
 then you will also have to manually configure the `redirectUri` to be `/`.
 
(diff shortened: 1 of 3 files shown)


bantic:
thanks again, @sukima !

bantic:
released in v0.9.5

PR #375, opened 2017-05-25
implement condition which checks if FB was already initialized and assigned to the window object

### LINKED ISSUE #296: Error when FB is already assigned to the window object

bantic:
via @andorfermichael

fixes #296

### DIFF
--- a/addon/providers/facebook-connect.js
+++ b/addon/providers/facebook-connect.js
@@ -1,8 +1,10 @@
 /* global FB, $ */
 
 /**
- * This class implements authentication against facebook
- * connect using the Facebook SDK.
+ * This class implements OAuth authorization for Facebook
+ * connect using the Facebook Connect SDK.
+ * It eagerly loads the Facebook Connect script using the settings provided
+ * unless it detects the presence of the global `window.FB`.
  */
 
 import Provider from 'torii/providers/base';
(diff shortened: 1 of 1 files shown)


bantic:
@andorfermichael Thanks for your PR! (#297) — I took your commit here and added a note in the comments explaining that existing `settings` would be ignored. If needed, that could be addressed in a later PR.

bantic:
released in v0.9.5

bantic:
released in v0.9.5

PR #376, opened 2017-05-25
URI-decode the OAuth2 authorization code

### LINKED ISSUE #323: URL decoding values in authData
Hey 👋

I'm currently working on a salesforce torii provider (current progress is here: https://github.com/whatthewhat/torii-salesforce-oauth2) based on the `oauth2-code` base provider.
Salesforce returns a URL encoded authorization code that needs to be decoded before using it to obtain an access token.. *I'm not willing to admit how much time it took me to figure out that this was the reason the authentication was failing* 😆

Basically I'm wondering if you think it makes sense for torii to call `decodeURIComponent` on all values in the authentication hash by default (would be happy to work on a PR for that) or if it's more of a per-provider concern and I should override `open` in the SF provider to achieve this?

whatthewhat:
#### Why is this needed?

Some providers (ahem, salesforce) return authorization codes that need to be URI-decoded (salesforce ones tend to have a '==' at the end). Plus it's probably a good idea to `decodeURIComponent` strings that come from a URL anyway.

Closes #323

### DIFF
--- a/addon/providers/oauth2-code.js
+++ b/addon/providers/oauth2-code.js
@@ -168,7 +168,7 @@ var Oauth2 = Provider.extend({
       }
 
       return {
-        authorizationCode: authData[responseType],
+        authorizationCode: decodeURIComponent(authData[responseType]),
         provider: name,
         redirectUri: redirectUri
       };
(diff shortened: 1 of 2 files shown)


bantic:
@whatthewhat Thanks! Looks good.

PR #377, opened 2017-05-30
Add DOCTYPE and character set declaration to redirect.html

treznick:
While totally not necessary for this behavior to work, adding a DOCTYPE and a `<meta charset>` tag is somewhat of a best practice. Happy to split this into two commits if that makes this easier. :)


mixonic:
Seems great. Thank you @treznick .

PR #381, opened 2017-08-03
[WIP] Allow incremental scopes for Google

YoranBrondsema:
In our application we need to request additional permissions for certain features (see https://developers.google.com/identity/protocols/OAuth2WebServer#incrementalAuth). Therefore, we need to be able to modify the `scope` parameter at "runtime" (when calling `this.get('torii').open`) rather than it being fixed from the configuration.

This PR contains a functional solution (it includes a working test) but is meant as a proof-of-concept to determine:

1. Whether this is something that should be supported in torii (I say yes :-))
2. Whether this is the right approach

Curious to see what you think!


YoranBrondsema:
@bantic Right now we're using our own fork of `torii` to use this functionality inside our application. This is not ideal so when you have time to look at this, your feedback about this PR will be very welcome!

Subtletree:
@mixonic Any chance this would get looked at if it gets rebased on master?

PR #383, opened 2017-08-18
Don't warn on empty config when included by an addon

ef4:
I'm making addons that depend on torii. They handle configuring torii for you. It would be helpful to make this warning go away for that scenario.


bantic:
Great, thanks @ef4 !

PR #385, opened 2017-09-01
use new modules api

Dhaulagiri:
Went smoothly with the exception of two imports that the codemod doesn't know about:

* `Ember.RouterDSL` - It seems like this could be added to the dataset used to polyfill back to the old globals imports as [`@ember/routing` includes several other similar modules](https://github.com/ember-cli/ember-rfc176-data#emberrouting)
* `Ember._ProxyMixin` - This is private and therefore my understanding is it won't be included in the new modules api, but I'm not sure what this means in practice here

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,13 +4,17 @@ module.exports = {
     ecmaVersion: 2017,
     sourceType: 'module'
   },
-  extends: 'eslint:recommended',
+  extends: [
+    'eslint:recommended',
+    'plugin:ember/recommended'
+  ],
   env: {
     browser: true
   },
   rules: {
+    'ember/no-global-jquery': 0, // this rule is broken
+    'ember/avoid-leaking-state-in-ember-objects': 0
   },
   globals: {
-    "Ember": false
   }
 };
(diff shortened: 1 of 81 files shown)


bantic:
@Dhaulagiri You're a saint.

Have you smoke-tested this build against any running apps?

I don't remember offhand what `Ember._ProxyMixin` does, but I am sure there's a refactor to get rid of it from the torii session service. 
Probably the same for `Ember.RouterDSL` although that might need more finesse.

Dhaulagiri:
It's a little tricky testing our torii setup locally, but I'll try to do that to at least sanity test that things are working as expected.

With respect to the two modules we are talking about, at this point there nothing requiring that the new modules imports be used so there is time to refactor these use cases.

Dhaulagiri:
Ha, well...our app didn't boot at first. The reason was that I accidentally also converted files in this addon's `/app` directory which is a no-no with this since we can't control how an app is going to transpile files in that directory (in fact our app is still using ember-cli-babel 5).  Once I undid that change things worked as expected.

Dhaulagiri:
@bantic got a bit sidetracked with real life but I did try this in our app and authing with a 3rd party service does work as it did before.

(Also included a commit to update us to ember-cli 2.16 here as well)

mixonic:
@Dhaulagiri this looks great :-) Thank you for taking this on.

PR #392, opened 2017-11-09
Make naming of testing modules consistent with Ember.js conventions

YoranBrondsema:
### Why
* Right now it's using a mix of e.g. `Unit - XXX` and `XXX - Unit`. This makes it hard to get an overview when viewing the tests in the browser.
* Using the Ember.js conventions makes it slightly easier for everyone I guess.

### What
e.g.

````
module('Popup - Unit')
````

becomes

````
module('Unit | Service | Popup')
````


bantic:
👏 thank you @YoranBrondsema !

PR #399, opened 2018-01-29
Stop using jQuery

cibernox:


addon/mixins/ui-service-mixin.js
```diff
@@ -45,7 +44,16 @@ var ServicesMixin = Mixin.create({
       }
 
       var remoteId = service.remoteIdGenerator.generate();
-
+      service.storageToriiEventHandler = function(storageEvent) {
```
mixonic: Seems like this would be better simply as a variable closed over for the  `addEventListener` and `removeEventListener` calls. Setting it onto the singleton service`` just pollutes the object.

addon/providers/facebook-connect.js
```diff
@@ -8,13 +8,17 @@
...
 
 import { run } from '@ember/runloop';
-
 import { Promise as EmberPromise } from 'rsvp';
-import { on } from '@ember/object/evented';
-
 import Provider from 'torii/providers/base';
 import { configurable } from 'torii/configuration';
 
+self._loadFacebookConnectScript = function(src) {
```
mixonic: Oof, could this instead be a utility file?

```js
let alternativeGetScript = false;

export function implementGetScript(state) {
  alternativeGetScript = state;
}

export function resetGetScript() {
  alternativeGetScript = false;
}

export function getScript(src) {
  if (alternativeGetScript) { alternativeGetScript(src); }
  let scriptTag = document.createElement('script');
  let firstScriptTag = document.getElementsByTagName('script')[0];
  scriptTag.src = src;
  firstScriptTag.parentNode.insertBefore(scriptTag, firstScriptTag);
}
```

Then you can use the set and reset helper in tests. I'd prefer this over stashing the var on the window.

cibernox:
@mixonic I've applied the feedback and also replaced `$.ajax` with a raw xhr.

cibernox:
I just installed my fork in an app and callbacks seem to be working fine.

addon/providers/-private/utils.js
```diff
@@ -0,0 +1,17 @@
...
+
+export function overrideLoadFacebookConnectScript(state) {
+  alternativeLoadFacebookConnectScript = state;
+}
+
+export function resetLoadFacebookConnectScript() {
+  alternativeLoadFacebookConnectScript = false;
+}
+
+export function loadFacebookConnectScript(src) {
```
mixonic: This is a general purpose `getScript` helper, there isn't anything specific to Facebook here. Am I missing it?

addon/providers/google-oauth2-bearer-v2.js
```diff
@@ -79,55 +79,46 @@ var GoogleOauth2BearerV2 = OAuth2Code.extend({
...
           }
-        );
-      }).then( function (authenticationData) {
+          */
+          // the token is valid if the 'audience' is the same as the
+          // 'client_id' (apiKey) provided to initiate authentication
+          if (jsonResponse.audience === clientId) {
+            // authentication succeeded. Add name and redirectUri to the
+            // authentication data and resolve
+            resolve(assign(authData, { provider: name, redirectUri: redirectUri }));
```
mixonic: `resolve` here and the `reject` calls below should be wrapped in `run`.

```js
run(() => {
  resolver(assign(authData, { provider: name, redirectUri }));
});
```

If they are not, they will cause an auto-run to be scheduled. I can see they weren't wrapped before, can we make that change anyway?

config/ember-try.js
```diff
@@ -97,7 +103,8 @@ module.exports = {
       },
       npm: {
         devDependencies: {
-          'ember-source': null
+          'ember-source': null,
```
mixonic: Ember beta is 3.0 which is not published to bower (I think? Canary wasn't last I checked). Addons generated by 3.0-beta use a  new format for ember-try config that pulls in ember builds for NPM. For example: https://github.com/ember-cli/ember-resolver/blob/d3d97643a81e325e703531792ef23be0a61bf48c/config/ember-try.js

mixonic: Dang, the bower build passed so I guess we are publishing to Bower still, @rwjblue if you see this, is that right? I'm surprised since I had to change to the new style to fix ember-resolver tests.

rwjblue: ember-source hasn't **deleted** the canary/beta/release branches in the bower repo yet, so grabbing them via bower will continue to work (yet the actual contents in those branches are pretty old...)...

mixonic: lol oh that is, like, even worse than what I thought was happening.

0xadada:
This would be a killer feature, i'm removing jQuery from my app here, and ran into the problem of a event listener using `$(window)` in `ui-service-mixin` here https://github.com/mirai-audio/mir/issues/31#issuecomment-362935370

addon/providers/-private/utils.js
```diff
@@ -0,0 +1,17 @@
...
+export function overrideLoadScript(state) {
+  alternativeLoadScript = state;
+}
+
+export function resetLoadScript() {
+  alternativeLoadScript = false;
+}
+
+export function loadScript(src) {
+  if (alternativeLoadScript) { alternativeLoadScript(src); }
```
mixonic: This should return, or the script injection will still occur

addon/providers/google-oauth2-bearer-v2.js
```diff
@@ -79,55 +80,41 @@ var GoogleOauth2BearerV2 = OAuth2Code.extend({
...
+          }
+        };
+        xhr.onerror = function() {
+          // authentication failed because the validation request failed
+          run(() => reject(new Error(`Token validation request failed with status '${xhr.statusText}' (server '${tokenValidationUrl}' '${xhr.responseText}').`)));
+        };
+        xhr.open('GET', `${tokenValidationUrl}?access_token=${encodeURIComponent(authData['access_token'])}`);
+        xhr.send();
+      }).then(function (authenticationData) {
         return authenticationData;
```
mixonic: Hm this `then` block seems unnecessary.

cibernox:
@mixonic Applied

0xadada:
new release?

mixonic:
@0xadada I had to do some cleanup (https://github.com/Vestorly/torii/pull/401), but it should now be released in 0.10.1 🎉

PR #420, opened 2018-08-08
Replace Edmodo Connect provider with addon torii-edmodo-connect

YoranBrondsema:
Because it operates in a niche market (K-12 education), I think the Edmodo Connect provider is not very popular. To do everyone a service and save off a few bytes in the `torii` addon, we extracted the provider into a third-party [torii-edmodo-connect](https://github.com/sutori/torii-edmodo-connect) addon.

This PR:
1. removes all code related to Edmodo Connect from the addon
2. adds a link to the third-party addon in the README


mixonic:
@YoranBrondsema that looks great. I wish we had an org for this stuff. Spinning out the individual packages and linking to them from here is a good first step!

PR #427, opened 2019-01-10
Replace merge with assign

Exelord:
As merge is deprecated I replaced it with recommended `assign` polyfill


rwwagner90:
@mixonic any chance we could merge this? If we need to support older Ember, we could add ember-assign-polyfill as well.

rwwagner90:
I opened a new PR adding ember-assign-polyfill on top of this

PR #429, opened 2019-01-13
handle merge deprecation

jdaviderb:
image

### DIFF
--- a/addon/services/popup.js
+++ b/addon/services/popup.js
@@ -1,7 +1,9 @@
 import Evented from '@ember/object/evented';
 import EmberObject from '@ember/object';
-import { merge } from '@ember/polyfills';
+import { merge, assign as emberAssign } from '@ember/polyfills';
 import UiServiceMixin from 'torii/mixins/ui-service-mixin';
+
+const assign = Object.assign || emberAssign || merge;
 
 function stringifyOptions(options){
   var optionsStrings = [];
@@ -29,7 +31,7 @@ function stringifyOptions(options){
 function prepareOptions(options){
   var width = options.width || 500,
       height = options.height || 500;
-  return merge({
+  return assign({
     left: ((screen.width / 2) - (width / 2)),
     top: ((screen.height / 2) - (height / 2)),
     width: width,
--- a/tests/helpers/start-app.js
+++ b/tests/helpers/start-app.js
@@ -1,12 +1,14 @@
 import Application from '../../app';
 import config from '../../config/environment';
-import { merge } from '@ember/polyfills';
+import { merge, assign as emberAssign } from '@ember/polyfills';
 import { run } from '@ember/runloop';
 
+const assign = Object.assign || emberAssign || merge;
+
 export default function startApp(attrs) {
-  let attributes = merge({}, config.APP);
+  let attributes = assign({}, config.APP);
   attributes.autoboot = true;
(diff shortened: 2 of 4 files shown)


tests/acceptance/routing-test.js
```diff
@@ -166,12 +166,11 @@ test('session.attemptedTransition is set before redirecting away from authentica
...
 
-  var applicationRoute = lookup(app, 'route:application');
+  var applicationRoute = lookup(app, 'route:secret');
+
   applicationRoute.reopen({
-    actions: {
-      accessDenied() {
-        attemptedTransition = this.get('session').attemptedTransition;
-      }
+    accessDenied() {
```
kategengler: Not sure this test change will reflect real usage. Per the documentation, `accessDenied` is put by the user into the `actions` hash.

kategengler:
This PR does not address #431 (easily verified by pointing the reproduction to the PR's branch).

jdaviderb:
@kategengler you're right, I added new changes in this PR, now it's working fine. Sorry for the push force I did 😢

rwwagner90:
@kategengler @jdaviderb could we use https://github.com/shipshapecode/ember-assign-polyfill and make the only change here switching from merge to assign?

addon/providers/oauth2-code.js
```diff
@@ -75,16 +75,9 @@ var Oauth2 = Provider.extend({
 
   state:        configurable('state', function () { return this.get('randomState'); }),
 
-  _randomState: null,
+  _randomState: Object.freeze(randomUrlSafe(16)),
```
mixonic: Hm, I think in the initial implementation this was intended to be lazy. I'm not sure why the private caching variable is ever needed though :thinking: Seems like this code was just poorly written in its original form?

rwwagner90:
@mixonic I think we should fix the merge deprecation separately from these other changes. We can use ember-assign-polyfill and do nothing but switch merge to assign.

jdaviderb:
@rwwagner90 I can do that and verify if the tests pass. @mixonic what do you think?

rwwagner90:
@jdaviderb I would go ahead and do that and move the other changes to another PR. It should be easy to get this reviewed and merged, if it's literally just switching merge to assign and adding the polyfill.

addon/providers/oauth2-code.js
jdaviderb: @mixonic I did this change to avoid mutate this property on a computed
https://github.com/Vestorly/torii/pull/429/files/2847792feb701e423ef114087e29d3bae835f45c#diff-5c9477407813ef428852917a5d1b8801L83

jdaviderb:
@rwwagner90 the problem is that without this change https://github.com/Vestorly/torii/pull/429/files#diff-a2f99fdbf3bf583f3e671163af305c07R41

`ember try` fail in ember 3.7 and we can not do merge because the tests fail in that version 

you can check it here

https://circleci.com/gh/Vestorly/torii/680?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

rwwagner90:
@jdaviderb that has nothing to do with converting assign to merge. If the tests are passing with merge right now, they should pass with assign and the polyfill.

rwwagner90:
@jdaviderb #427 actually does the swap. Perhaps we should get that merged in?

PR #434, opened 2019-05-13
Dropping unused module

ef4:
This doesn't appear to be used, and it breaks embroider.

(That is, it breaks embroider when embroider has `staticAddonTrees` off to emulate traditional ember-cli, because traditional ember-cli will definitely include this unused file into the build, and we try to follow along, but fail because the file refers to nonexistent imports. When `staticAddonTrees` is on, embroider ignores this file successfully since nothing uses it.)


mixonic:
I think this goes back to when the project was used via Bower: https://github.com/Vestorly/torii/commit/0ebffceff25cc8aaab9f506c4c62681117bbd57c#diff-04c6e90faac2675aa89e2176d2eec7d8L270

And can be removed.

